Issue #3, opened 2023-06-22
Combine bounty related commands into /bounty


ntseng:
Using sub-commands, bounty related commands can be organized to be shown together. Relevant commands include:
- [x] #2 - /bounty post
- [ ] #4 - /bounty edit
- [ ] #5 - /bounty add-completer
- [ ] #6 - /bounty remove-completer
- [ ] #7 - /bounty complete
- [ ] #8 - /bounty take-down

ntseng:
Subcommand structure is set up as of #2

PR #23, opened 2023-07-04
#2: port `/bounty post`

### LINKED ISSUE #2: Port /post
This slash command allows users to post bounties. It will require making the Bounty database model. A modal will be a good UI for this instead of the text parsing we were doing before.

ntseng:
Summary
-------
- ported `/bounty post`
   - includes slash command, bounty slot select, and input modal
- added models for Bounty, GuildRank, and Completion
   - GuildRank now uses a composite primary key of guildId + varianceThreshold
   - Completion is the combination of PendingCompleter and BountyCompletion
- `/create-bounty-board` now populates with posts for existing bounties
- `/create-bounty-board` now creates a forum in ListView as default (shows bounty posts better than GalleryView)
- split `hunter.isRankEligible` from `hunter.rank` (now `INTEGER | NULL` from `STRING | NULL`)

Local Tests Performed
---------------------
- [x] tested bounty post flow, excluding placement calculation due to lack of way to increment other users' XP

Issue
-----
Closes #2

### DIFF
--- a/source/commands/bounty.js
+++ b/source/commands/bounty.js
@@ -1,0 +1,65 @@
+const { PermissionFlagsBits, ActionRowBuilder, StringSelectMenuBuilder } = require('discord.js');
+const { CommandWrapper } = require('../classes');
+const { database } = require('../../database');
+const { getNumberEmoji } = require('../helpers');
+
+const customId = "bounty";
+const options = [];
+const subcommands = [
+	{
+		name: "post",
+		description: "Post your own bounty (+1 XP)",
+		optionsInput: []
+	}
+];
+module.exports = new CommandWrapper(customId, "Bounties are user-created objectives for other server members to complete", PermissionFlagsBits.ViewChannel, false, true, 3000, options, subcommands,
+	(interaction) => {
+		switch (interaction.options.getSubcommand()) {
+			case subcommands[0].name: // Post
+				database.models.Hunter.findOne({ where: { userId: interaction.user.id, guildId: interaction.guildId } }).then(async hunter => {
+					if (!hunter) { //TODO use findOrCreate after double-checking associations (may not cascade from hunter to user correctly)
+						const user = await database.models.User.findByPk(interaction.user.id);
+						if (!user) {
+							await database.models.User.create({ id: interaction.user.id });
+						}
+						hunter = await database.models.Hunter.create({ userId: interaction.user.id, guildId: interaction.guildId, isRankEligible: !interaction.member.manageable });
+					}
+
+					const { maxSimBounties } = await database.models.Guild.findByPk(interaction.guildId);
+
+					const existingBounties = await database.models.Bounty.findAll({ where: { userId: interaction.user.id, guildId: interaction.guildId, state: "open" } });
+					const occupiedSlots = existingBounties.map(bounty => bounty.slotNumber);
+					const bountySlots = hunter.maxSlots(maxSimBounties);
+					const slotOptions = [];
+					for (let slotNumber = 1; slotNumber <= bountySlots; slotNumber++) {
+						if (!occupiedSlots.includes(slotNumber)) {
+							slotOptions.push({
+								emoji: getNumberEmoji(slotNumber),
+								label: `Slot ${slotNumber}`,
+								description: `Reward: ${hunter.slotWorth(slotNumber)} XP`,
+								value: slotNumber.toString()
+							})
+						}
+					}
+
+					if (slotOptions.length > 0) {
+						interaction.reply({
+							content: "You can post a bounty for other server members to help out with. Here's some examples:\n\t• __Party Up__ Get bounty hunters to join you for a game session\n\t• __WTB/WTS__ Get the word out your looking to trade\n\t• __Achievement Get__ Get help working toward an achievement\n\nTo make a bounty, you'll need:\n\t• a title\n\t• a description\nOptionally, you can also add:\n\t• a url for an image\n\t• a start and end time (to make an event to go with your bounty)\n\nKeep in mind that while you're in charge of adding completers and ending the bounty, the bounty is still subject to server rules and moderation.",
+							components: [
+								new ActionRowBuilder().addComponents(
+									new StringSelectMenuBuilder().setCustomId("bountypostselect")
+										.setPlaceholder("XP awarded depends on slot used...")
+										.setMaxValues(1)
+										.setOptions(slotOptions)
+								)
+							],
+							ephemeral: true
+						});
+					} else {
+						interaction.reply({ content: "You don't seem to have any open bounty slots at the moment.", ephemeral: true });
+					}
+				})
+				break;
+		}
+	}
+);
--- a/source/commands/bounty.js
+++ b/source/commands/bounty.js
@@ -12,7 +12,7 @@ const subcommands = [
 		optionsInput: []
 	}
 ];
-module.exports = new CommandWrapper(customId, "Bounties are user-created objectives for other server members to complete", PermissionFlagsBits.ViewChannel, false, true, 3000, options, subcommands,
+module.exports = new CommandWrapper(customId, "Bounties are user-created objectives for other server members to complete", PermissionFlagsBits.ViewChannel, false, false, 3000, options, subcommands,
 	(interaction) => {
 		switch (interaction.options.getSubcommand()) {
 			case subcommands[0].name: // Post
--- a/source/commands/create-bounty-board.js
+++ b/source/commands/create-bounty-board.js
@@ -44,7 +44,7 @@ module.exports = new CommandWrapper(customId, "Create a new bounty board forum c
 			scoreboardPost.pin();
 			guildProfile.scoreId = scoreboardPost.id;
 
-			database.models.Bounty.findAll({ where: { guildId: interaction.guildId, state: "open" } }).then(bounties => {
+			database.models.Bounty.findAll({ where: { guildId: interaction.guildId, state: "open" }, order: [["createdAt", "DESC"]] }).then(bounties => {
 				for (const bounty of bounties) {
 					bounty.asEmbed(interaction.guild, undefined, guildProfile).then(bountyEmbed => {
 						return bountyBoard.threads.create({
--- a/source/commands/create-bounty-board.js
+++ b/source/commands/create-bounty-board.js
@@ -27,7 +27,7 @@ module.exports = new CommandWrapper(customId, "Create a new bounty board forum c
 			],
 			//TODO use "availableTags" to allow tagging bounties ("completed", "event", "open" as default tags?)
 			defaultSortOrder: SortOrderType.CreationDate,
-			defaultForumLayout: ForumLayoutType.GalleryView,
+			defaultForumLayout: ForumLayoutType.ListView,
 			reason: `/create-bounty-board by ${interaction.user}`
 		}).then(async bountyBoard => {
 			let guildProfile = await database.models.Guild.findByPk(interaction.guildId);
@@ -44,7 +44,19 @@ module.exports = new CommandWrapper(customId, "Create a new bounty board forum c
 			scoreboardPost.pin();
 			guildProfile.scoreId = scoreboardPost.id;
 
-			//TODO create posts for existing bounties
+			database.models.Bounty.findAll({ where: { guildId: interaction.guildId, state: "open" } }).then(bounties => {
+				for (const bounty of bounties) {
+					bounty.asEmbed(interaction.guild, undefined, guildProfile).then(bountyEmbed => {
+						return bountyBoard.threads.create({
+							name: bounty.title,
+							message: { embeds: [bountyEmbed] }
+						})
+					}).then(posting => {
+						bounty.postingId = posting.id;
+						bounty.save()
+					})
+				}
+			});
 
 			guildProfile.save();
 			interaction.reply({ content: `A new bounty board has been created: ${bountyBoard}`, ephemeral: true });
--- a/source/commands/stats.js
+++ b/source/commands/stats.js
@@ -1,4 +1,4 @@
-const { PermissionFlagsBits, EmbedBuilder } = require('discord.js');
+const { EmbedBuilder } = require('discord.js');
 const { CommandWrapper } = require('../classes');
 const { buildGuildStatsEmbed, randomFooterTip, ihpAuthorPayload } = require('../embedHelpers');
 const { database } = require('../../database');
--- a/source/commands/stats.js
+++ b/source/commands/stats.js
@@ -10,7 +10,7 @@ const options = [
 	{
 		type: "User",
 		name: "bounty-hunter",
-		description: "Whose stats to check, BountyBot for the guild's stats",
+		description: "Whose stats to check (defaults to yourself); input BountyBot for the server stats",
 		required: false,
 		choices: []
 	}
--- a/source/commands/stats.js
+++ b/source/commands/stats.js
@@ -16,7 +16,7 @@ const options = [
 	}
 ];
 const subcommands = [];
-module.exports = new CommandWrapper(customId, "Get the BountyBot stats for yourself or someone else", PermissionFlagsBits.ViewChannel, false, false, 3000, options, subcommands,
+module.exports = new CommandWrapper(customId, "Get the BountyBot stats for yourself or someone else", null, false, false, 3000, options, subcommands,
 	/** Get the BountyBot stats for yourself or someone else */
 	(interaction) => {
 		const target = interaction.options.getMember("user"); //TODO switch from magic string "user" to options[0].name
--- a/source/commands/stats.js
+++ b/source/commands/stats.js
@@ -10,7 +10,7 @@ const options = [
 	{
 		type: "User",
 		name: "bounty-hunter",
-		description: "Whose stats to check (defaults to yourself); input BountyBot for the server stats",
+		description: "Whose stats to check; BountyBot for the server stats, empty for yourself",
 		required: false,
 		choices: []
 	}
--- a/source/constants.js
+++ b/source/constants.js
@@ -1,5 +1,6 @@
 exports.SAFE_DELIMITER = "→";
 exports.MAX_SET_TIMEOUT = 2 ** 31 - 1;
+exports.YEAR_IN_MS = 31556926000;
 
 exports.authPath = "../config/auth.json";
 const { testGuildId, feedbackChannelId } = require(exports.authPath);
--- a/source/helpers.js
+++ b/source/helpers.js
@@ -118,7 +118,7 @@ exports.setRanks = async (participants, ranks) => {
  * @returns an array of rank and placement update strings
  */
 exports.getRankUpdates = async function (guild, force = false) {
-	const allHunters = await database.models.Hunter.findAll({ where: { guildId: guild.id, seasonXP: { [Op.gt]: 0 } }, order: [["seasonXP", "DESC"]] });
+	const allHunters = await database.models.Hunter.findAll({ where: { guildId: guild.id }, order: [["seasonXP", "DESC"]] });
 	const ranks = await database.models.GuildRank.findAll({ where: { guildId: guild.id }, order: [["varianceThreshold", "ASC"]] });
 	return exports.setRanks(allHunters, ranks).then(async (firstPlaceMessage) => {
 		const roleIds = ranks.filter(rank => rank.roleId != "").map(rank => rank.roleId);
--- a/source/helpers.js
+++ b/source/helpers.js
@@ -140,10 +140,12 @@ exports.getRankUpdates = async function (guild, force = false) {
 				let destinationRole;
 				if (member.manageable) {
 					await member.roles.remove(roleIds);
-					const rankRoleId = ranks[hunter.rank].roleId;
-					if (rankRoleId) {
-						await member.roles.add(rankRoleId);
-						destinationRole = await guild.roles.fetch(rankRoleId);
+					if (hunter.isRankEligible) { // Feature: remove rank roles from DQ'd users but don't give them new ones
+						const rankRoleId = ranks[hunter.rank].roleId;
+						if (rankRoleId) {
+							await member.roles.add(rankRoleId);
+							destinationRole = await guild.roles.fetch(rankRoleId);
+						}
 					}
 				}
 				if (hunter.rank > hunter.lastRank) { // Feature: don't comment on rank downs
--- a/source/helpers.js
+++ b/source/helpers.js
@@ -79,9 +79,18 @@ exports.setRanks = async (participants, ranks) => {
 			hunter.nextRankXP = 0;
 		}
 	}
-	const n = Math.max(rankableHunters.length, 2);
-	mean /= n;
-	const stdDev = Math.sqrt(rankableHunters.reduce((total, hunter) => total + (hunter.seasonXP - mean) ** 2, 0) / n);
+
+	if (rankableHunters.length < 2) {
+		for (const hunter of participants) {
+			hunter.rank = null;
+			hunter.seasonPlacement = 0;
+			hunter.save();
+		}
+		return null;
+	}
+
+	mean /= rankableHunters.length;
+	const stdDev = Math.sqrt(rankableHunters.reduce((total, hunter) => total + (hunter.seasonXP - mean) ** 2, 0) / rankableHunters.length);
 	for (const hunter of rankableHunters) {
 		let variance = (hunter.seasonXP - mean) / stdDev;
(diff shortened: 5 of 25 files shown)


source/commands/create-bounty-board.js
```diff
@@ -44,7 +44,19 @@ module.exports = new CommandWrapper(customId, "Create a new bounty board forum c
 			scoreboardPost.pin();
 			guildProfile.scoreId = scoreboardPost.id;
 
-			//TODO create posts for existing bounties
+			database.models.Bounty.findAll({ where: { guildId: interaction.guildId, state: "open" } }).then(bounties => {
```
SillySalamndr, line 47: Very niche concern here: we should try to do some sorting here on the query to have a consistent order by the time we loop through and make bounty board posts.

source/commands/stats.js
```diff
@@ -10,7 +10,7 @@ const options = [
 	{
 		type: "User",
 		name: "bounty-hunter",
-		description: "Whose stats to check, BountyBot for the guild's stats",
+		description: "Whose stats to check (defaults to yourself); input BountyBot for the server stats",
```
SillySalamndr, line 13: This parenthetical feels...off. I think this would fit better after the semicolon right now. Maybe convert the semicolon to a period, and separate the default/BB bit by a semicolon?

source/constants.js
```diff
@@ -1,5 +1,6 @@
 exports.SAFE_DELIMITER = "→";
 exports.MAX_SET_TIMEOUT = 2 ** 31 - 1;
+exports.YEAR_IN_MS = 31556926000;
```
SillySalamndr, line 3: Are we sure that there isn't something within JS by default with this value, so we don't have to have a random constant for this?

source/constants.js
```diff
@@ -1,5 +1,6 @@
 exports.SAFE_DELIMITER = "→";
 exports.MAX_SET_TIMEOUT = 2 ** 31 - 1;
```
SillySalamndr, line 2: Does this match max size 32 bit integer? This *looks* like that, so JS may already have a constant for it.

source/commands/stats.js
ntseng: This `Whose stats to check. Input BountyBot for the server stats; defaults to yourself`?

I personally don't like periods at the end of sentence fragments. What about the following?

`Whose stats to check; input BountyBot for the server stats; none defaults to yourself`

source/constants.js
ntseng: Happy to use it if you can find it, I couldn't.

source/constants.js
ntseng: You might be thinking of MAX_SAFE_INTEGER, which is 2^53 - 1

source/helpers.js
```diff
@@ -15,3 +36,123 @@ exports.generateTextBar = function (numerator, denominator, barLength) {
...
+		case 8:
+			return '8️⃣';
+		case 9:
+			return '9️⃣';
+		case 10:
+			return '🔟';
+		default:
+			return '#️⃣';
+	}
+}
```
SillySalamndr, line 67: As much as this looks somewhat textbook for a switch case block, it also feels...inefficient? the 0 through 10 case could be pulled from a constant array that has each of those at the corresponding index, and that somewhat reduces this into an if...else (if num >=0 && numb <= 10 ...).

ntseng: Probably better as a hash table tbh

ntseng: Done in https://github.com/Imaginary-Horizons-Productions/BountyBot/pull/23/commits/0797fc93322413eb711e7ac461e854b3810b2cda

source/helpers.js
```diff
@@ -15,3 +36,123 @@ exports.generateTextBar = function (numerator, denominator, barLength) {
...
+				previousFirstPlaceId = hunter.userId;
+			}
+			hunter.lastRank = hunter.rank;
+			mean += hunter.seasonXP;
+			rankableHunters.push(hunter);
+		} else {
+			hunter.nextRankXP = 0;
+		}
+	}
+	const n = Math.max(rankableHunters.length, 2);
```
SillySalamndr, line 82: ...Does this fully make sense? What if there's only one rankable hunter? If this solves a weird edge case with exactly one hunter, I feel like that edge case would affect everything down the line to the point that we just want different handling for the only 1 hunter case, potentially skipping on ranks in the first place.

source/helpers.js
```diff
@@ -15,3 +36,123 @@ exports.generateTextBar = function (numerator, denominator, barLength) {
...
+	mean /= n;
+	const stdDev = Math.sqrt(rankableHunters.reduce((total, hunter) => total + (hunter.seasonXP - mean) ** 2, 0) / n);
```
SillySalamndr, line 84: An entirely reasonable use of reduce, unlike what tends to happen with it spiraling out of control. Good stuff!

source/helpers.js
```diff
@@ -15,3 +36,123 @@ exports.generateTextBar = function (numerator, denominator, barLength) {
...
+	for (const hunter of rankableHunters) {
+		let variance = (hunter.seasonXP - mean) / stdDev;
+		ranks.forEach((rank, index) => {
```
SillySalamndr, line 87: Is ranks always ordered as we want?

source/commands/create-bounty-board.js
ntseng: Added in https://github.com/Imaginary-Horizons-Productions/BountyBot/pull/23/commits/18e3edb67d7e1990d7baa5575428a74b3887fe5f

source/helpers.js
ntseng: Yep, they're ordered by varianceThreshold ascending in the fetch

source/helpers.js
```diff
@@ -15,3 +36,123 @@ exports.generateTextBar = function (numerator, denominator, barLength) {
...
+			if (variance >= rank.varianceThreshold) {
+				hunter.rank = index;
```
SillySalamndr, line 89: Still storing this by order index on the hunter seems...volatile? Now that we have a real database with relations, we could store a real reference to the rank in the DB instead.

source/commands/stats.js
ntseng: Here's how it actually shows in the UI
image
Maybe it should only be the special cases

source/helpers.js
ntseng: GuildRank uses a composite key of guildId and varianceThreshold. I guess we could decided on a hashing convention?

source/commands/bounty.js
```diff
@@ -0,0 +1,65 @@
...
+const customId = "bounty";
+const options = [];
+const subcommands = [
+	{
+		name: "post",
+		description: "Post your own bounty (+1 XP)",
+		optionsInput: []
+	}
+];
+module.exports = new CommandWrapper(customId, "Bounties are user-created objectives for other server members to complete", PermissionFlagsBits.ViewChannel, false, true, 3000, options, subcommands,
```
SillySalamndr, line 15: Something that I'm noticing with this wrapper functionality is that it currently looks/feels overloaded with parameters, which leads to run-on-like lines like this. Reading these parameters is pretty tough, especially without using an IDE that gets tailored to this use case. Probably not something to address in the immediate term, but something to look at improving in the future. In the more immediate future, figuring out clever ways to have better line breaks within this line may help with readability.

source/commands/bounty.js
```diff
@@ -0,0 +1,65 @@
...
+	(interaction) => {
+		switch (interaction.options.getSubcommand()) {
+			case subcommands[0].name: // Post
```
SillySalamndr, line 18: This looks a bit overengineered in the immediate term, but probably fine after adding a means to edit with this. Something probably better to mention though: it may be smart in the long term to have the functionality of each subcommand attached to it or otherwise put to a separate function, so that we can break out the code/functionality in a modular way.

source/commands/bounty.js
ntseng: image
In that case... cooldown, execute, and isPremium aren't used in the slash command builder, and thus would be good candidates for pulling out of the constructor. The inheritance structure would be pretty empty at that point too

source/commands/bounty.js
```diff
@@ -0,0 +1,65 @@
...
+				database.models.Hunter.findOne({ where: { userId: interaction.user.id, guildId: interaction.guildId } }).then(async hunter => {
```
SillySalamndr, line 19: This might benefit from a similar await scheme as the code contained within it. The anonymous function will need to be `async` then, but that should be as easy as prepending it with that keyword.

source/commands/bounty.js
ntseng: Planning to organize commands in UI with this structure (`/bounty post`, `/bounty edit`, `/bounty complete`, etc). See #3 for more info.

Totally fine with separating subcommands out into separate files or whatever since their responsibilities are going to be very different, but haven't gotten to it yet because that will also entail adapting the wiki generator script to be able to use those files.

source/commands/bounty.js
```diff
@@ -0,0 +1,65 @@
...
+					if (!hunter) { //TODO use findOrCreate after double-checking associations (may not cascade from hunter to user correctly)
+						const user = await database.models.User.findByPk(interaction.user.id);
+						if (!user) {
+							await database.models.User.create({ id: interaction.user.id });
+						}
+						hunter = await database.models.Hunter.create({ userId: interaction.user.id, guildId: interaction.guildId, isRankEligible: !interaction.member.manageable });
+					}
```
SillySalamndr, line 26: This code of make-hunter-if-doesn't-exist probably belongs in its own utility function, since it'd be also useful when adding completers, completing a bounty outright, or toasting. I'd say that belongs in the DAO-like layer somewhere, but that's beyond the scope of this.

source/commands/bounty.js
ntseng: I tend to default to .then and only use async/await if it simplifies flow to try and get away with not promisifying things unnecessarily.

source/models/users/User.js
```diff
@@ -1,7 +1,7 @@
 const { DataTypes: { STRING, BOOLEAN }, Model } = require('sequelize');
 
 /** This class stores global information for bot users */
-let userModel = {
+const userModel = {
```
SillySalamndr, line 4: Any particular reason for this switch?

source/commands/bounty.js
SillySalamndr: Well, in this case, it'll keep a lot of the same checks/comparisons in the same flow and should reduce indentation in a meaningful way. We know that this whole function needs to become asynchronous anyway for this check, so there's no harm in having it be async from the get go.

source/commands/bounty.js
ntseng: I agree! However, I think Sequelize's out-of-the-box `findOrCreate` is exactly what we're looking, so we won't need this in the final code anyways. Unfortunately, because Hunter implies a User and I don't think we have associations configured correctly, switching to `findOrCreate` is juuust out of scope for this PR.

source/models/users/User.js
ntseng: It would suck if `userModel` got reassigned accidentally, and this would get the terminal to comment if it does.

source/commands/bounty.js
SillySalamndr: I'm more concerned about putting them in separate functions first, since their functionality shouldn't directly interact with each other outside of having the conceptual relation of `Bounty`, which this file already accomplishes. Personally, I'd prefer putting all of the related bounty "controller" code in the same place if possible, and putting them in separate functions may allow us to bundle all `bounty` code from different UI sources in the future.

source/commands/bounty.js
SillySalamndr: Something something KISS rule? Probably an effort for a different pull request.

source/commands/stats.js
SillySalamndr: Yeah, I personally prefer periods on sentence fragments for semi-declarative stuff like this, but could see how it could be grating. I like that first one that you don't, but consistency is more important than anything else. I do like that second option from before the screenshots.

source/commands/bounty.js
ntseng: I'd be fine with that. A little skeptical about how frequently we'll actually get to reuse those functions with discord API tying interface to functionality, but if we can find some it'd be a win.

source/helpers.js
SillySalamndr: Personally, I'm not looking for a hashing convention. We *should* be able to have `Rank` directly on the `Hunter` object as a foreign key, and then `Hunter` will automagically have both foreign keys (if done right).

source/models/users/User.js
SillySalamndr: Yep, makes sense.

source/commands/bounty.js
ntseng: Yeah, Button, Select and Modal all fitting the same "non slash command" payload was a plesaent surprise

source/commands/bounty.js
ntseng: Yeah, I guess we probably can't get away with pretending we'll always acknowledge interactions in 3 seconds forever (and thus will have to use a deferred reply). At that point I'd prefer the execute signature to be `(interaction) => Promise<void>` than include the case where it's not returning a promise. So I guess something to note for when we do that class refactor.

source/helpers.js
ntseng: Sounds fancy!

source/models/users/Hunter.js
```diff
@@ -104,9 +113,50 @@ exports.Hunter = class Hunter extends Model {
...
+	}
+
+	/**
+	 * @param {Guild} guild
+	 * @param {number} points
+	 * @param {boolean} ignoreMultiplier
+	 * @returns {string} level-up text
+	 */
+	async addXP(guild, points, ignoreMultiplier) {
+		const guildProfile = await database.models.Guild.findByPk(guild.id);
```
SillySalamndr, line 127: Models having partial references to other models in a way that isn't that model owning an instance of the other is usually a pitfall that leads to circular references. Are we confident that we won't have Guild pulling for Hunters in the future?

source/commands/stats.js
ntseng: Done in https://github.com/Imaginary-Horizons-Productions/BountyBot/pull/23/commits/62e6aa7fe3b75ffafe650a6bd8e8e235c9dd859f

source/models/users/Hunter.js
ntseng: I have no certainty on that question. On the one hand, Guild level is derived from total Hunter level, but most uses of that doesn't live on the Guild entity? I mostly don't think we have enough pieces in play to make the call yet.

source/helpers.js
ntseng: Constraint: n = 1 causes a divide by 0 later, but we still want to touch everyone for the case of "first recalculation after a season reset".
Alleged previous thinking: I think we settled on this as a compromise because only one person being a season participant was something of a trivial case.

Haven't thought through what the functionality of a more formal special case would be though.

source/commands/bounty.js
SillySalamndr: It is much less about reuse and more about concentrating conceptually adjacent things. Having all `bounty` functionality in similar locations means that developers in the future have less overall preparation needed to do to work on `bounty` related functionality instead of bouncing between more contexts. In my opinion, bad context switching kills more of my time at work than many more serious issues, since it tends to happen incredibly frequently when it does occur.

SillySalamndr: Although, to reiterate again from earlier: that kind of work would likely be something that should be a cleanup pass later on, and shouldn't be a reason to hold this back right now.

source/models/users/Hunter.js
SillySalamndr: I don't know how to put this especially lightly, but not knowing where data belongs like this is an incredibly bad sign. The most important factor of a relational database is its constructed and structured relationships, and data that doesn't have proper relational setup should either be treated as if it is everywhere all the time (bad because it makes it harder to work on code by expanding context to things that don't actually belong) or as if it doesn't exist always (since the only reason why data has value in this scheme is that it has a relationship to some other data and some concept/table encapsulating it). So, if this information is to exist right now, we want to make sure it fits in with a relationship as it is created in the codebase.

For a recommendation: this data does not belong on the individual hunter at all, since it relates to multiple database entries at once and not any individual hunter. The most appropriate place in my opinion is on the Guild, since that is the primary object that relates all the relevant data for this data point, in spite of (or really because of?) it being calculated using child components in the data hierarchy. We really don't want a new column for this, but we don't have to use one, either. (Maybe a better way of putting it: Guilds own the list of Hunters, so Guilds can know about Hunters and their data directly, while Hunters do not own and are not responsible for Guild data.) In my opinion, this seems like a good opportunity for a `virtual` field on the Guild that calculates the value when it is requested - this way we don't even need to do the calculations here at all. https://sequelize.org/docs/v6/core-concepts/getters-setters-virtuals/#virtual-fields

SillySalamndr: (This is also a reason why most enterprise software has a DAO and/or Business Logic layer; it allows for disentangling the calculations from the data objects themselves.)

source/commands/bounty.js
SillySalamndr: There's a potential issue with `findOrCreate` where we want a `User` model created before the `Hunter` one due to the data structure and relationship between them, but in most places that are relevant we only really *need* the `Hunter`. This is something that we'll want to track as we implement `findOrCreate`, potentially abstracting this interaction into a separate function due to that relationship tie.

source/helpers.js
SillySalamndr: Eh, the hashing function thing is the fancier bit, imo.

source/models/bounties/Bounty.js
```diff
@@ -0,0 +1,124 @@
...
+const { Guild: HunterGuild } = require('../guilds/Guild');
+const { database } = require('../../../database');
+
+/** Bounties are user created objectives for other server members to complete */
+const bountyModel = {
+	id: {
+		primaryKey: true,
+		type: BIGINT,
+		autoIncrement: true
+	},
```
SillySalamndr, line 14: I know that we've talked about this at some point, but is there a reasonable future where the primary key is the combination of the user id, guild id, and slot number? that may be a bit much to have a 3-part primary key, especially if we persist this information in the database after completion, though.

source/models/bounties/Bounty.js
```diff
@@ -0,0 +1,124 @@
...
+	title: {
+		type: STRING,
+		allowNull: false,
+		defaultValue: "\u200B"
+	},
+	description: {
+		type: STRING,
+		allowNull: false,
+		defaultValue: "\u200B"
```
SillySalamndr, line 48: Do these need to be zero-width spaces?

source/models/bounties/Bounty.js
```diff
@@ -0,0 +1,124 @@
...
+	 * @param {Hunter?} hunter
+	 * @param {HunterGuild?} hunterGuild
+	 */
+	asEmbed(guild, hunter, hunterGuild) {
+		return guild.members.fetch(this.userId).then(async author => {
+			if (!hunter) {
+				hunter = await database.models.Hunter.findOne({ where: { userId: this.userId, guildId: this.guildId } });
+			}
+			if (!hunterGuild) {
+				hunterGuild = await database.models.Guild.findByPk(this.guildId);
```
SillySalamndr, line 87: Since the Bounty doesn't really "own" the hunter or guild, I'd probably just prefer if we always pass the hunter and hunterGuild parameters and throw an error in these spots, instead.

source/models/bounties/Bounty.js
```diff
@@ -0,0 +1,124 @@
...
+			}
+
+			const embed = new EmbedBuilder().setColor(author.displayColor)
+				.setAuthor(ihpAuthorPayload)
```
SillySalamndr, line 91: This is likely due to me not being in the zone of discord.js for some time, but the `ihp` abbreviation feels a little too abstract, and I have no real idea of what this refers to. Is this `Imaginary Horizons Productions`? If so, that also feels too broad for this bot, and `bb` for `BountyBot` would make more sense in context.

source/modals/bountypostmodal.js
```diff
@@ -0,0 +1,112 @@
...
+
+				if (endTimestamp >= nowTimestamp + (5 * YEAR_IN_MS)) {
+					errors.push("End timestamp cannot be 5 years in the future or further.");
+				}
+			}
+		}
+
+		if (errors.length > 0) {
+			interaction.reply({ content: `The following errors were encountered while posting your bounty **${title}**:\n• ${errors.join("\n• ")}`, ephemeral: true });
+		} else {
```
SillySalamndr, line 68: To avoid needing an explicit `else` block, it would be better to just return in the `if` block here, since it also ends the validation steps.

source/modals/bountypostmodal.js
```diff
@@ -0,0 +1,112 @@
...
+					scheduledStartTime: startTimestamp * 1000,
+					scheduledEndTime: endTimestamp * 1000,
+					privacyLevel: 2,
+					entityType: GuildScheduledEventEntityType.External,
+					entityMetadata: { location: `${interaction.member.displayName}'s #${slotNumber} Bounty` }
+				};
+				if (imageURL) {
+					eventPayload.image = imageURL;
+				}
+				const event = await interaction.guild.scheduledEvents.create(eventPayload);
```
SillySalamndr, line 87: Is there a reason to save the promise results for the event here? We don't end up using the returned value to save/attach information to the Bounty itself, so this just...hangs there for garbage collection. It may be wise to at minimum `try..catch` this to tell the user if the event creation had an error that we didn't manage to catch with validation.

source/models/users/Hunter.js
SillySalamndr: Okay, to make the transition to having `guildXP` be virtual, I *think* this is all that you'll really need:
```
//...
guildXP: {
   type: DataTypes.VIRTUAL,
   async get() {
      return sequelize.query(`SELECT FLOOR(SQRT(SUM(level)/(:xpCoefficient + 1))) FROM hunters WHERE guildId = :guildId`, {
         replacements: [{ xpCoefficient: this.xpCoefficient }, { guildId: this.id }],
         type: QueryTypes.SELECT
      });
   }
}
//...
```

SillySalamndr: Additional relevant docs:

- https://sequelize.org/docs/v6/core-concepts/raw-queries/
- https://sequelize.org/api/v6/class/src/sequelize.js~sequelize#instance-method-query

SillySalamndr (changes requested):
A couple changes between `Hunter` and `Bounty` and I think this is good to go. Those changes are mentioned on the relevant files, and there are some additional things that may or may not be relevant that I've also commented on.

SillySalamndr (approved):

source/models/bounties/Bounty.js
ntseng: If we went with a composite key for Bounties, that means only current bounties would be allowed in the Bounties table (which would also mean we'd want another table for `completed|deleted` bounties). I think it's be more likely we'll want to have both current and historic bounties in the same table.

source/models/bounties/Bounty.js
ntseng: The constant is the data required to make the circled portion of an embed. I think the name is appropriate.
Screenshot 2023-07-14 130427

source/modals/bountypostmodal.js
ntseng: The `event.id` is used in L88.

source/modals/bountypostmodal.js
ntseng: Fixed in https://github.com/Imaginary-Horizons-Productions/BountyBot/pull/23/commits/1539b761523841a8428a3e214bc2ef5bbcf964cb

source/models/bounties/Bounty.js
ntseng: Removed in https://github.com/Imaginary-Horizons-Productions/BountyBot/pull/23/commits/d01fbba08ed572cf8539025c696959361cd08b91

source/commands/bounty.js
ntseng: IIRC, setting up associations correctly will include "Make a User if trying to make a Hunter and it's missing."

source/models/bounties/Bounty.js
ntseng: I consider this temporary (maybe I should have commented as so?), with a future investigation being "can we always pass the hunter and hunterGuild?" to be done when everything is implemented. In the "can always pass" world, I would probably even skip out on error throwing.

PR #254, opened 2024-10-03
SillySalamndr/issue186

### LINKED ISSUE #186: Add shortcut to /stats as User Context Menu
This will also include updating main interaction handling to support context menu interactions

SillySalamndr:
Summary
-------
- Added new command types for context menu commands
- Added handling for the new context menu data that mirrors similar commands
- Added a command to pull stats for a given user

Local Tests Performed
---------------------
- [X] Bot still turns on (no build errors, circular dependencies, or start-up errors)
- [X] Displayed that the context menu option appears when right clicking on a user
- [X] Demonstrated that the context menu option has appropriate responses

Issue
-----
Closes #186


SillySalamndr (commented):
Something to note for a future task: I think that the argument list for each of these interaction wrappers is complex enough with enough potentially optional/should have defaults arguments that using the object wrapped arguments pattern may be advisable. Not something that needs to be done immediately or ever, though.

source/context_menus/_context_menu_stats.js
```diff
@@ -0,0 +1,58 @@
+const { PermissionFlagBits, InteractionContextType } = require('discord.js');
+const { UserContextMenuWrapper } = require('../classes');
+
+const mainId = "User Stats";
```
ntseng, line 4: Please update `wiki/Developers.md` with the context for this exception for id.

source/context_menus/_context_menu_stats.js
```diff
@@ -0,0 +1,58 @@
...
+module.exports = new UserContextMenuWrapper(mainId, null, false, [ InteractionContextType.Guild ], 3000,
+	/** Specs */
+	(interaction, database, runMode) => {
+		const target = interaction.targetMember;
+		if (target.id == interaction.client.user.id) {
+			// BountyBot
+			buildCompanyStatsEmbed(interaction.guild, database).then(embed => {
```
ntseng, line 11: Missing import for `buildCompanyStatsEmbed()`

source/context_menus/_context_menu_stats.js
```diff
@@ -0,0 +1,58 @@
...
+								{ name: "Total XP Earned", value: `${hunter.xp} XP`, inline: true },
+								{ name: "Most Seconded Toast", value: mostSecondedToast ? `"${mostSecondedToast.text}" with **${mostSecondedToast.secondings} secondings**` : "No toasts seconded yet..." },
+								{ name: "Bounty Stats", value: `Bounties Hunted: ${hunter.othersFinished} bount${hunter.othersFinished === 1 ? 'y' : 'ies'}\nBounty Postings: ${hunter.mineFinished} bount${hunter.mineFinished === 1 ? 'y' : 'ies'}`, inline: true },
+								{ name: "Toast Stats", value: `Toasts Raised: ${hunter.toastsRaised} toast${hunter.toastsRaised === 1 ? "" : "s"}\nToasts Seconded: ${hunter.toastsSeconded} toast${hunter.toastsSeconded === 1 ? "" : "s"}\nToasts Recieved: ${hunter.toastsReceived} toast${hunter.toastsReceived === 1 ? "" : "s"}`, inline: true },
+							)
+							.setFooter(randomFooterTip())
```
ntseng, line 50: Missing import for `randomFooterTip()`

source/context_menus/_context_menu_stats.js
```diff
@@ -0,0 +1,58 @@
...
+				const rankName = ranks[hunter.rank]?.roleId ? `<@&${ranks[hunter.rank].roleId}>` : `Rank ${hunter.rank + 1}`;
+				const mostSecondedToast = await database.models.Toast.findOne({ where: { senderId: target.id, companyId: interaction.guildId, secondings: { [Op.gt]: 0 } }, order: [["secondings", "DESC"]] })
+
+				interaction.reply({
+					embeds: [
+						new EmbedBuilder().setColor(Colors[hunter.profileColor])
+							.setAuthor(ihpAuthorPayload)
+							.setThumbnail(target.user.avatarURL())
+							.setTitle(`${target.displayName} is __Level ${hunter.level}__`)
+							.setDescription(`${generateTextBar(hunter.xp - currentLevelThreshold, nextLevelThreshold - currentLevelThreshold, 11)}\nThey have earned *${currentParticipation?.xp ?? 0} XP* this season${hunter.rank !== null ? ` which qualifies for ${rankName}` : ""}.`)
```
ntseng, line 42: Missing import for `generateTextBar()`

source/context_menus/_context_menu_stats.js
```diff
@@ -0,0 +1,58 @@
...
+		if (target.id == interaction.client.user.id) {
+			// BountyBot
+			buildCompanyStatsEmbed(interaction.guild, database).then(embed => {
+				interaction.reply({
+					embeds: [embed],
+					ephemeral: true
+				});
+			})
+		} else {
+			// Other Hunter
```
ntseng, line 18: Please create a function for this and the own stats embeds so that functionality can stay in step with the slash command.

ntseng (changes requested):
A few missing imports and a request to create some util functions.

source/context_menus/_context_menu_stats.js
ntseng, line 4: "BountyBot Stats" or "Profile Stats" might be more clear which context the stats are from.

SillySalamndr: As much as I'm leery about beginning to attach "BountyBot" to all our context menus, I think "Profile" is the same level of confusing as "User" here. "BountyBot" it is.

ntseng: I'm also not a fan of an unconditional "BountyBot" prefix (I don't think it'll end up bleeding into too many more, but I guess we'll see on that one?).

source/bot.js
```diff
@@ -111,6 +112,20 @@ client.on(Events.InteractionCreate, interaction => {
 			const choices = unfilteredChoices.filter(choice => choice.value.toLowerCase().includes(focusedOption.value.toLowerCase()))
 				.slice(0, 25);
 			interaction.respond(choices);
+		} else if (interaction.isContextMenuCommand()) {
+			const contextMenu = getContextMenu(interaction.commandName);
+			if (contextMenu.premiumCommand && !premium.paid.includes(interaction.user.id) && !premium.gift.includes(interaction.user.id)) {
+				interaction.reply({ content: `The \`/${interaction.commandName}\` command is a premium command. Learn more with ${commandMention("premium")}.`, ephemeral: true });
```
ntseng, line 118:
```suggestion
				interaction.reply({ content: `The \`/${interaction.commandName}\` context menu option is a premium feature. Learn more with ${commandMention("premium")}.`, ephemeral: true });
```
Wording update for accuracy

source/bot.js
```diff
@@ -111,6 +112,20 @@ client.on(Events.InteractionCreate, interaction => {
...
+				return;
+			}
+
+			const cooldownTimestamp = contextMenu.getCooldownTimestamp(interaction.user.id, interactionCooldowns);
+			if (cooldownTimestamp) {
+				interaction.reply({ content: `Please wait, the \`/${interaction.commandName}\` command is on cooldown. It can be used again <t:${cooldownTimestamp}:R>.`, ephemeral: true });
```
ntseng, line 124:
```suggestion
				interaction.reply({ content: `Please wait, the \`/${interaction.commandName}\` context menu option is on cooldown. It can be used again <t:${cooldownTimestamp}:R>.`, ephemeral: true });
```
Wording update for accuracy here too

source/classes/InteractionWrapper.js
```diff
@@ -135,9 +135,72 @@ class SelectWrapper extends InteractionWrapper {
...
+		this.builder = new ContextMenuCommandBuilder()
+			.setName(mainIdInput)
+			.setDescription(descriptionInput)
+			.setDMPermission(allowInDMsInput)
+			.setType();
+		if (defaultMemberPermission) {
+			this.builder.setDefaultMemberPermissions(defaultMemberPermission);
+		}
+	}
+};
```
ntseng, line 150: Were these changes to `SelectWrapper` intended? The empty `.setType()` and other interface changes seem like they might not function properly.

source/context_menus/_contextMenuDictionary.js
```diff
@@ -0,0 +1,26 @@
+const { ContextMenuWrapper } = require("../classes");
+
+/** @type {string[]} */
+exports.contextMenuFiles = [
+	"_context_menu_stats.js"
```
ntseng, line 5: File name convention questions:
- Was there a reason for the leading underscore in the stats filename? I've been using that to sort "important" files to the top of the directory
- Confirming: the intent was to keep both user context menu options and message context menu options in the same dictionary, accepting the uniqueness constraint created within those two categories?

source/context_menus/_context_menu_stats.js
```diff
@@ -0,0 +1,58 @@
...
+			database.models.Hunter.findOne({ where: { userId: target.id, companyId: interaction.guildId } }).then(async hunter => {
+				if (!hunter) {
+					interaction.reply({ content: "The specified user doesn't seem to have a profile with this server's BountyBot yet. It'll be created when they gain XP.", ephemeral: true });
+					return;
+				}
+
+				const { xpCoefficient } = await database.models.Company.findByPk(interaction.guildId);
+				const currentLevelThreshold = Hunter.xpThreshold(hunter.level, xpCoefficient);
```
ntseng, line 26: Got a crash on `Hunter` not being imported here (for it trying to use the static method).

source/context_menus/_context_menu_stats.js
```diff
@@ -0,0 +1,58 @@
+const { PermissionFlagBits, InteractionContextType } = require('discord.js');
```
ntseng, line 1:
```suggestion
const { InteractionContextType } = require('discord.js');
```
PermissionFlagBits is unused

source/context_menus/_context_menu_stats.js
SillySalamndr: That...shouldn't be the case? If we imported hunter here, it shouldn't matter, since we're exclusively accessing Hunter via the database model list, and the imported Hunter object (not fed into the Sequelize functionality) likely wouldn't have (and shouldn't have) that function in the first place, at least at the line that I *think* you are pointing at. If you are pointing to the ones below this, that isn't clear by how you've placed this comment.

ntseng: It's the `Hunter.xpThreshold()` static method that's the issue. I agree it's weird that github seems to know Hunter's a class, but the crash was found while running the command on the branch.

source/classes/InteractionWrapper.js
SillySalamndr: They were not, and I have changes ready to fix this soon.

source/context_menus/_context_menu_stats.js
SillySalamndr: Should be up to date now!

source/context_menus/_context_menu_stats.js
SillySalamndr: This will be moved to its own tasking

ntseng (approved):

PR #256, opened 2024-10-06
#222: add context menu Grant Bounty Credit

### LINKED ISSUE #222: Context menu: add completer on user
Opens a modal to accept a bounty slot number, then adds the clicked user as a completer

ntseng:
Summary
-------
- add context menu Grant Bounty Credit
   - move business logic to `source/engines/bountyEngine.js` to ensure same function between bounty board button, slash command, and context menu sources
   - future optimization of bounty board button not fetching own thread channel available

Local Tests Performed
---------------------
- [x] adding completer via context menu is shown in updated bounty post on bounty board
- [x] no regressions detected after refactor of bounty board button and slash command sources of adding completers

Issue
-----
Closes #222


source/engines/bountyEngine.js
```diff
@@ -0,0 +1,35 @@
...
+			return bountyBoard.threads.fetch(bounty.postingId);
+		}).then(posting => {
+			posting.send({ content: `${listifyEN(completerIds.map(id => userMention(id)))} ${completerIds.length === 1 ? "has" : "have"} been added as ${completerIds.length === 1 ? "a completer" : "completers"} of this bounty! ${congratulationBuilder()}!` });
+		});
+	}
+}
+
+module.exports = {
+	addCompleters
+}
```
SillySalamndr, line 35: I'm not personally convinced by the "engines" naming convention for something like this. I'd personally expect engines to be fully self-sufficient (pure input to output) and closer to utility files with more internalized complexity, but the way that they are being used here is inherently not self-sufficient (needs a database object in order to function). I'm creating a `logic` folder instead, with the file name matching the simplest iteration on the associated functionality (so `stats.js` in my implemented case). What's your opinion on this?

ntseng: I don't have strong opinions on the name. I like the overall structure, but `stats.js` by itself is a little more narrow than my first instinct. Maybe groupings similar to our database models (Users, Seasons, Toasts, etc) would consolidate things intuitively?

source/context_menus/_context_menu_stats.js
SillySalamndr, line 1: Should be removed.

SillySalamndr (approved):

Issue #257, opened 2024-10-06
Optimization: Skip fetching own thread when Adding Completers via bounty board button


ntseng:
The Add Completers button is only created within their own bounty board threads, thus the interactions generated by them have the reference to their thread included. The current implementation doesn't account for this and unnecessarily requests the thread from dAPI instead.

PR #301, opened 2025-01-20
Silly salamndr/issue264

### LINKED ISSUE #264: Move `posting.send` in `addCompleter` to controller layer
The motivation for this effort is to improve our frontend context sensitivity. One payoff for this is it'll allow us to more easily implement the following:
> Responses to adding completers via slash command or context menu are non-ephemeral by default (when no bounty board exists) so that the response can double as a public notification, but become ephemeral when a bounty board exists because the update is going to be tracked in the thread. Button logic doesn't need this contingency because no buttons will exist when a bounty board doesn't exist.

This increased context sensitivity is also where the issue ties in with #257, since the button interaction gets the reference to the thread for free.

SillySalamndr:
Summary
-------
The work on this task is a bit split between the guessed/assumed goal (setting up the basis for future logic/command/util code separation) and the real goal (moving some code to the command layer so that the re-duplicated code can be specialized reasonably). Unfortunately, not all of the work of the former goal is quite done, but it is set up in a way that should make that effort much faster in the future regardless of it feeling incomplete and the latter goal is set up for future specialization in a hopefully reasonable way.

Local Tests Performed
---------------------
- [X] bot is able to add completers using all 3 methods

Issue
-----
Closes #264


ntseng:
Also fixes #257

PR #306, opened 2025-02-04
#258: refactor `raiseToast()` to use logic layer

### LINKED ISSUE #258: Move `interaction.reply` from `raiseToast` to controllers
Acknowledging interactions has been deemed controller responsibility rather than engine responsibility. Further, there are minor optimizations possible with knowing if `toasteeIds` is guaranteed to have a single element (when coming from the context menu) or if array handling is needed.

ntseng:
Summary
-------
- created logic files for companies and hunters; added database logic to toasts
- created logic functions `findOrCreateCompany`, `findOrCreateBountyHunter`, and `findOneHunter` for encapsulating database transactions to logic layer
   - implemented in toasts flows only to control scope
- moved interaction acknowledgement and reward message generation from logic layer to frontend
- fixed the return type on `Hunter.addXP()` and added type annotation to logic files
- performance optimizations for `raiseToast`
- added "sender is not banned" validation to context menu flow

Future Work
-----
1. Decouple `progressGoal` and `raiseToast` so that embeds can be generated in frontend instead of `raiseToast`
2. Implement logic functions throughout rest of codebase
3. Consolidate reward message generation and button component generation (and embed generation after 1.) in frontend to avoid behavior desyncs
   - Switch formatting in reward message generation to discord.js formatting functions

Local Tests Performed
---------------------
- [x] context menu and slash command flows for toasts complete without crash or text error
- [x] critical toast still provides XP message

Issue
-----
Closes #258


source/context_menus/Raise_a_Toast.js
```diff
@@ -42,7 +53,46 @@ module.exports = new UserContextMenuWrapper(mainId, PermissionFlagsBits.SendMess
...
+					if (rankUpdates.length > 0) {
+						content += `\n\n__**Rank Ups**__\n- ${rankUpdates.join("\n- ")}`;
+					}
+					if (rewardTexts.length > 0) {
+						content += `\n\n__**Rewards**__\n- ${rewardTexts.join("\n- ")}`;
+					}
+					if (content.length > MAX_MESSAGE_CONTENT_LENGTH) {
+						content = `Message overflow! Many people (?) probably gained many things (?). Use ${commandMention("stats")} to look things up.`;
+					}
+				}
```
SillySalamndr, line 83: This content string generation code is repeated a couple of times in ways that I'd expect we'd want to keep in sync, but not necessarily frequently enough that there's a huge drive to ensure that we're consolidating this work at the moment. Are we expecting that we'll have another means to raise a toast? (If so, a task to consolidate this in a `util`-like file would be good to get ahead of that future need; otherwise, I think it is a coin toss for whether it is worth it to do this type of refactor.) Are we expecting the text generation to diverge in these locations?

source/logic/toasts.js
```diff
@@ -1,33 +1,41 @@
...
-async function raiseToast(interaction, database, toasteeIds, toastText, imageURL = null) {
-	const [company] = await database.models.Company.findOrCreate({ where: { id: interaction.guildId } });
+async function raiseToast(guild, company, sender, senderHunter, toasteeIds, toastText, imageURL = null) {
 	const embeds = [
 		new EmbedBuilder().setColor("e5b271")
 			.setThumbnail(company.toastThumbnailURL ?? 'https://cdn.discordapp.com/attachments/545684759276421120/751876927723143178/glass-celebration.png')
 			.setTitle(toastText)
 			.setDescription(`A toast to ${listifyEN(toasteeIds.map(id => userMention(id)))}!`)
-			.setFooter({ text: interaction.member.displayName, iconURL: interaction.user.avatarURL() })
+			.setFooter({ text: sender.displayName, iconURL: sender.user.avatarURL() })
```
SillySalamndr, line 31: I know that this isn't entirely the fault/changes from this issue, but if we're going to push for the separation of our logic and Discord interface layers, we'll want to move this embed building to the Discord interface file(s).

SillySalamndr (commented):
Two potential problems, but I'm not sure if we want to address either of them here with this issue. In one of those problems (content string generation), I'm not sure we want to address it at all.

source/logic/toasts.js
ntseng: I also noticed this, but felt that disentangling it would be too much scope (because it'd require moving the later Goal stuff too). This is a good reminder to make a new issue.

source/context_menus/Raise_a_Toast.js
ntseng: This is shared among all places that provide XP, including both "toast by command" and "toast by context menu". I'll make a new issue to consolidate to `Toast.generateRewardMessage()` (and separate ones for `Bounty.generateRewardMessage()`, and `Seconding.generateRewardMessage()`).

ntseng: New issues: #318, #319, & #320

source/logic/toasts.js
ntseng: New issue: #321

SillySalamndr (approved):

PR #307, opened 2025-02-07
#298: replace uses of `Bounty.asEmbed()` with `Bounty.embed()`

### LINKED ISSUE #298: Refactor `asEmbed` on Bounty (and places that use it) to no longer require database access
Initial work has been done on this with `embed` on Bounty being a version of `asEmbed` that does not require the database, but many locations are still using `asEmbed` and should find ways to use `embed` instead.

ntseng:
Summary
-------
- replace uses of `Bounty.asEmbed()` with `Bounty.embed()`
- fix type annotation on `updateBoardPosting` functions' parameters
- added dev mode skip to bounty board showcase button
- fix race condition between interaction acknowledgement and ephemeral message clean up in Bounty Board Add Completers
- fix evergreen completions clobbering old records of completions on `xpAwarded`
- fix bounty completions sending "+0 GP" reward messages
- evergreen bounty embeds now create a completers field for completions
- fixed an `undefined` in `/evergreen swap` response message
- evergreen bounties now contribute GP (for each completer)
- fix server goal progress bars sometimes finding the wrong goal
- creates consolidated function for finding current goal GP
- removes `database` argument from `progressGoal`

Local Tests Performed
---------------------
- no regressions in the following flows:
   - [x] Bounty Board Add Completers button
   - [x] Bounty Board Complete button
   - [x] Bounty Board Remove Completers button
   - [x] /bounty add-completers
   - [x] /bounty complete
   - [x] /bounty edit
   - [x] /bounty list
   - [x] /bounty post
   - [x] /create-default bounty-board forum (with existing bounties)
   - [x] /evergreen complete
   - [x] /evergreen edit
   - [x] /evergreen post
   - [x] /evergreen showcase
   - [x] /evergreen swap
   - [x] /evergreen take-down
   - [x] Context Menu: Give Bounty Credit
   - [x] /bounty showcase
   - [x] Bounty Board Showcase button
   - [x] Bonus Bounty Showcase item

Issue
-----
Closes #298

### DIFF
--- a/source/commands/evergreen/complete.js
+++ b/source/commands/evergreen/complete.js
@@ -86,7 +86,7 @@ async function executeSubcommand(interaction, database, runMode, ...args) {
 		}
 	}
 
-	bounty.asEmbed(interaction.guild, company.level, company.festivalMultiplierString(), true, database).then(embed => {
+	bounty.embed(interaction.guild, company.level, true, company, await database.models.Completion.findAll({ where: { bountyId: bounty.id } })).then(embed => {
 		return interaction.reply({ embeds: [embed], withResponse: true });
 	}).then(response => {
 		getRankUpdates(interaction.guild, database).then(rankUpdates => {
--- a/source/commands/evergreen/complete.js
+++ b/source/commands/evergreen/complete.js
@@ -59,19 +59,18 @@ async function executeSubcommand(interaction, database, runMode, ...args) {
 	season.increment("bountiesCompleted");
 
 	const rawCompletions = [];
+	// Evergreen bounties are not eligible for showcase bonuses
+	const bountyBaseValue = Bounty.calculateCompleterReward(company.level, slotNumber, 0);
+	const bountyValue = bountyBaseValue * company.festivalMultiplier;
 	for (const userId of dedupedCompleterIds) {
 		rawCompletions.push({
 			bountyId: bounty.id,
 			userId,
-			companyId: interaction.guildId
+			companyId: interaction.guildId,
+			xpAwarded: bountyValue
 		});
 	}
-	await database.models.Completion.bulkCreate(rawCompletions);
-
-	// Evergreen bounties are not eligible for showcase bonuses
-	const bountyBaseValue = Bounty.calculateCompleterReward(company.level, slotNumber, 0);
-	const bountyValue = bountyBaseValue * company.festivalMultiplier;
-	database.models.Completion.update({ xpAwarded: bountyValue }, { where: { bountyId: bounty.id } });
+	const completions = await database.models.Completion.bulkCreate(rawCompletions);
 
 	for (const userId of validatedCompleterIds) {
 		const hunter = await database.models.Hunter.findOne({ where: { companyId: interaction.guildId, userId } });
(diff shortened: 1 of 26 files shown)


source/commands/evergreen/complete.js
```diff
@@ -59,35 +57,55 @@ async function executeSubcommand(interaction, database, runMode, ...args) {
...
+		contributorIds.forEach(id => finalContributorIds.add(id));
 	}
 
-	bounty.asEmbed(interaction.guild, company.level, company.festivalMultiplierString(), true, database).then(embed => {
-		return interaction.reply({ embeds: [embed], withResponse: true });
+	bounty.embed(interaction.guild, company.level, true, company, completions).then(async embed => {
+		const acknowledgeOptions = { embeds: [embed], withResponse: true };
+		if (totalGP > 0) {
+			levelTexts.push(`This bounty contributed ${totalGP} GP to the Server Goal!`);
+			const [goal] = await database.models.Goal.findAll({ where: { companyId: interaction.guildId, state: wasGoalCompleted ? "completed" : "ongoing" }, order: [["createdAt", "DESC"]], limit: 1 });
```
SillySalamndr, line 95: I've already added a review comment for a similar location to this where a findAll is used to find one record and the same potential issues exist that we should guard against here, with the possibility of a single pure text query as well.

SillySalamndr (changes requested):
Most everything looks good, but I'm concerned about a couple of locations where there's a query that uses findAll where only 1 record is intended to be found. In those cases, we'll want to make sure that we're handling the cases where more than 1 or no records are returned so we don't accidentally fall into more severe errors in the future in these locations. Notably, all of these locations are immediately following up into another query, so we may have better time using a string query to fold both of these queries into one more complex but faster operation.

source/commands/evergreen/complete.js
SillySalamndr: I know that you believe that this is resolved as-is, but this is still in bad form. This query that intends to find one result either needs to be a `findOne` to match that intent or needs to confirm after the query that one result was found.

source/logic/toasts.js
```diff
@@ -74,7 +74,7 @@ async function raiseToast(interaction, database, toasteeIds, toastText, imageURL
 			);
 		}
 		if (progressData.gpContributed > 0) {
-			const goal = await database.models.Goal.findOne({ where: { companyId: interaction.guildId } });
+			const [goal] = await database.models.Goal.findAll({ where: { companyId: interaction.guildId, state: progressData.goalCompleted ? "completed" : "ongoing" }, order: [["createdAt", "DESC"]], limit: 1 });
```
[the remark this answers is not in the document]
SillySalamndr: I know that you believe that this is resolved as-is, but this is still in bad form. This query that intends to find one result either needs to be a findOne to match that intent or needs to confirm after the query that one result was found.

SillySalamndr (changes requested):

source/logic/toasts.js
```diff
@@ -74,9 +74,8 @@ async function raiseToast(interaction, database, toasteeIds, toastText, imageURL
 			);
 		}
 		if (progressData.gpContributed > 0) {
-			const goal = await database.models.Goal.findOne({ where: { companyId: interaction.guildId } });
-			const progress = await database.models.Contribution.sum("value", { where: { goalId: goal.id } }) ?? 0;
-			embeds[0].addFields({ name: "Server Goal", value: `${generateTextBar(progress, goal.requiredContributions, 15)} ${Math.min(progress, goal.requiredContributions)}/${goal.requiredContributions} GP` });
+			const { currentGP, requiredGP } = await findLatestGoalProgress(interaction.guildId);
```
SillySalamndr, line 77: This is good!

SillySalamndr:
The preview was...wrong? It should be good now.

SillySalamndr (approved):

PR #316, opened 2025-02-14
#233: support sequelize-cli for migration generators

### LINKED ISSUE #233: Refactor migrations pattern to use Sequelize generators
https://sequelize.org/docs/v6/other-topics/migrations/

ntseng:
Summary
-------
- support sequelize-cli for migration generators
- prune some unused model associations

Local Tests Performed
---------------------
- [x] used `/bounty list` to fetch bounties from database
- [x] used `/toast` to add and update rows in database
- [x] confirmed consolidated migration properly updates database

Issue
-----
Closes #233


config_template/config.json
```diff
@@ -0,0 +1,15 @@
...
+		"storage": "./db.sqlite"
+	},
+	"test": {
+		"dialect": "sqlite",
+		"storage": "./db.sqlite"
+	},
+	"production": {
+		"dialect": "sqlite",
+		"storage": "./db.sqlite",
+		"logging": false
```
SillySalamndr, line 13: good change here

migrations/20250214064057-pre-233.js
SillySalamndr, line 1: This file name is messy and doesn't really convey any meaning to developers other than a timestamp of when it was generated (which is less meaningful than desirable). Is it reasonable/possible to have a more meaningful name here?

source/bot.js
```diff
@@ -40,16 +40,13 @@ const client = new Client({
...
-const databasePromise = connectToDatabase(runMode).then(database => {
-	for (logicFile in logicBlob) { // Set the database for the logic files that store it
-		logicBlob[logicFile].setDB?.(database);
-	}
-	setCommandLogic(logicBlob);
-	setButtonLogic(logicBlob);
-	setSelectLogic(logicBlob);
-	setContextMenuLogic(logicBlob);
-	return database; // Pass the database through
-});
```
SillySalamndr, line 52: This change looks like a regression. Something that the old function did was establish the connection to the database using the `authenticate` function, guaranteeing that we had a database connection before we tried to set up these associations, which we are no longer doing.

migrations/20250214064057-pre-233.js
ntseng: Timestamp portion of the name is automatically generated via the scripts. I personally like it because it's unambiguous about what order the migrations should be run in.

source/bot.js
```diff
@@ -111,71 +108,67 @@ client.on(Events.ClientReady, () => {
 });
 
 client.on(Events.InteractionCreate, interaction => {
-	databasePromise.then(database => {
```
SillySalamndr, line 114: Same as above here. We're no longer checking that we have a database connection, which we want to guarantee before we do any actions that can touch it.

source/bot.js
```diff
@@ -193,9 +186,7 @@ client.on(Events.ChannelDelete, channel => {
 });
 
 client.on(Events.MessageDelete, message => {
-	databasePromise.then(database => {
```
SillySalamndr, line 196: Same as above here. We're no longer checking that we have a database connection, which we want to guarantee before we do any actions that can touch it.

source/bot.js
```diff
@@ -204,9 +195,7 @@ client.on(Events.MessageDelete, message => {
 });
 
 client.on(Events.ThreadDelete, thread => {
-	databasePromise.then(database => {
```
SillySalamndr, line 207: Same as above here. We're no longer checking that we have a database connection, which we want to guarantee before we do any actions that can touch it.

source/bot.js
```diff
@@ -215,24 +204,22 @@ client.on(Events.ThreadDelete, thread => {
 })
 
 client.on(Events.GuildDelete, guild => {
-	databasePromise.then(database => {
```
SillySalamndr, line 218: Same as above here. We're no longer checking that we have a database connection, which we want to guarantee before we do any actions that can touch it.

SillySalamndr (commented):
LGTM

PR #322, opened 2025-02-16
#299: Move user ID validation from the command file to the relevant logic file for adding completers

### LINKED ISSUE #299: Move user ID validation from the command file to the relevant logic file for adding completers
Moving ID validation to the logic file will mean that the command files will no longer need the database passed to them, which will support our effort in separating our discord-facing concerns from our database/logic-facing concerns. We won't be able to remove the parameter yet do to other commands also passing that parameter, but it is one step in the right direction for that effort.

Incredibly tiny risk present here: only 2 of the 3 code routes that are relevant to adding completers do ID validation. This will likely add ID validation to the database on the remaining code path, which will reduce performance on that specific call. Fortunately, that call is already the easiest/fastest call, so I don't expect any real issues here, outside of the obvious ones that would already exist from doing a refactor.

SillySalamndr:
Summary
-------
Moved user validation to the logic layer.
Moved bounty validation to the logic layer.
Only validation that can happen without database knowledge now happens on the command layer.

Some additional work and polish was done to make sure that these changes behave sanely, like forcing ceratin datatypes to be arrays to prevent errors and having better error catching from the logic layer in our command layer files.

Local Tests Performed
---------------------
- [X] bot is able to give credit in all 3 main ways (slash command, button on bounty board, context menu)
- [X] bot provides relevant error messages when it cannot add users as completers

Issue
-----
Closes #299


source/logic/bounties.js
```diff
@@ -21,11 +21,48 @@ function setDB(database) {
  * @param {Guild} guild
  * @param {Bounty} bounty
  * @param {Company} company
- * @param {string[]} completerIds
+ * @param {GuildMember[]} completerMembers
+ * @param {string} runMode
  */
-async function addCompleters(guild, bounty, completerIds) {
+async function addCompleters({slotNumber, posterId, bountyId}, guild, completerMembers, runMode) {
```
ntseng, line 27: The type of the first parameter of this function (`{slotNumber, posterId, bountyId}`) is not documented. I would also prefer its type be `{ slotNumber: number, posterId: string } | bountyId: string | Bounty`, so that the type narrows later as it gets checked.

source/buttons/bbaddcompleters.js
```diff
@@ -50,32 +44,20 @@ module.exports = new ButtonWrapper(mainId, 3000,
...
-					if (hunter.isBanned) {
-						bannedIds.push(memberId);
-						continue;
+				try {
+					let { bounty: returnedBounty, allCompleters, poster, company, validatedCompleterIds } = await addCompleters({bountyId, posterId: interaction.user.id}, collectedInteraction.guild, Array.from(collectedInteraction.members.values()), runMode);
+					updateBoardPosting(returnedBounty, company, poster, validatedCompleterIds, allCompleters, collectedInteraction.guild, interaction.channel);
+					return collectedInteraction.update({
+						components: []
+					});
+				} catch (e) {
```
ntseng, line 53: My first instinct would not be to mix our error messages with ones coming from dAPI and differentiating on type. What do you like about that pattern?

ntseng (changes requested):
Overall super happy to see completer id validation get consolidated and moved to the logic layer.

Have some suggestions that merit discussion before merge, especially one about keeping behavior the same.

source/logic/bounties.js
```diff
@@ -21,11 +21,48 @@ function setDB(database) {
...
+	// Validate completer IDs
+	const validatedCompleterIds = [];
+	const existingCompletions = await db.models.Completion.findAll({ where: { bountyId: bounty.id, companyId: guild.id } });
+	const existingCompleterIds = existingCompletions.map(completion => completion.userId);
+	const bannedIds = [];
+	for (const member of completerMembers.filter(member => !existingCompleterIds.includes(member.id))) {
+		if (runMode === "prod" && member.user.bot) continue;
+		const memberId = member.id;
+		await db.models.User.findOrCreate({ where: { id: memberId } });
+		const [hunter] = await db.models.Hunter.findOrCreate({ where: { userId: memberId, companyId: guild.id } });
```
ntseng, line 51:
```suggestion
		const [hunter] = await findOrCreateBountyHunter(memberId, guild.id);
```
Note `findOrCreateBountyHunter()` will need to be imported.

SillySalamndr: I'll get to this one soon-ish.

source/buttons/bbaddcompleters.js
SillySalamndr: I'm a bit confused on this. Isn't that what I'm already doing with line 54 and 56? Our errors that are intended to be user facing are always "raw" strings, while errors thrown by Discord.JS are always an otherwise more specific error type. (I came to this code as it currently stands due to the crossover that was already happening during testing, and it was sufficient to be able to have Discord.JS errors appear in the console while our user message strings would appear in Discord itself.)

source/buttons/bbaddcompleters.js
```diff
@@ -32,12 +32,6 @@ async function updateBoardPosting(bounty, company, poster, newCompleterIds, comp
 const mainId = "bbaddcompleters";
 module.exports = new ButtonWrapper(mainId, 3000,
 	(interaction, [bountyId], database, runMode) => {
-		database.models.Bounty.findByPk(bountyId, { include: database.models.Bounty.Company }).then(async bounty => {
```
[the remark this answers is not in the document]
SillySalamndr: I'll admit that the way that I refactored this was informed by patterns that I'm used to in Java EJB styling, where each time you move between the command/controller and the logic/DAO layers, you incur a pretty hefty performance penalty (as the Java beans do some cleanup and synchronizing with the database each time you do so). We do not have any of the downsides of that system here, and doing what you're suggesting is reasonable. It shouldn't bee _too_ much effort, but if this starts to get out of hand, I'll move that effort to a new ticket.

source/logic/bounties.js
SillySalamndr: This will be replaced by simply passing the whole bounty due to the linked refactor https://github.com/Imaginary-Horizons-Productions/BountyBot/pull/322#discussion_r1957516707

source/buttons/bbaddcompleters.js
SillySalamndr: I think I'm starting to partially understand your actual statement here. The oddity here is that if we try to handle this "normally", we'd be returning a string instead of the object that we do destructured assignment on above, which would make that process a bit more awkward (but not impossible). Additionally, these messages do behave like errors, just as ones that we tell the user, so being able to "return" them via throwing an error means that we can split the output types (data object blob and error message) more cleanly.

ntseng: The above comment was an attempt to start a discussion, apologies if it read as critical. The way I see it, `addCompleters()` has three types of things it can return: a data blob on success, a string on failure, and an Error on dAPI error. I think it's fair to group the string and the Error cases, since they both represent failure states; mostly wanted to think through the design of it. Is it better to group the blob and string because they're "expected" outcomes and we want unexpected Error handling to be consolidated elsewhere? Maybe we shouldn't be grouping at all and just `switch` through the returned types individually.

source/logic/bounties.js
```diff
@@ -18,27 +19,26 @@ function setDB(database) {
 }
 
 /**
- * @param {Guild} guild
+ * @param {{slotNumber: number, posterId: string} | string} bountyInfo
+ * @param {Guild | undefined} guild
```
ntseng, line 23: Looks like `guild` is only being used in the non-`bountyId` case to provide the `companyId`. I think it'd be cleaner to include `companyId` in that case of `bountyInfo` instead.

source/logic/bounties.js
```diff
@@ -18,27 +19,26 @@ function setDB(database) {
...
+ * @returns {Bounty | null} the bounty, if it exists
```
ntseng, line 24: This type can be abbreviated to `Bounty?`. I'm still on the fence about whether I like that better though.

source/buttons/bbaddcompleters.js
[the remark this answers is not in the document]
ntseng: Admittedly, this has the button flow doing a W shape in our layers (which I vaguely remember being considered poor form?). Not totally sure how to avoid that with two validation points and a needed response to the frontend though...

source/buttons/bbaddcompleters.js
SillySalamndr: What really matters here is whatever we end up doing, we do it uniformly. We really want to get the logic layer refactor done for the dev QoL, so maybe we delay this discussion? Maybe it is the next big discussion point? Idk.

source/logic/bounties.js
SillySalamndr: I'm going to do this for now, but I'm not sure I like how the JSDoc becomes critical to the understanding of this function's parameters, which wasn't an issue with the previous destructuring (where all parameter names were visible in the function declaration, even if the were not all used). Not something that I'm going to try to work out now, but it feels like JSDoc is too critical to being able to read and understand the code at a baseline level.

source/logic/bounties.js
```diff
@@ -18,14 +19,48 @@ function setDB(database) {
...
+		return await db.models.Bounty.findByPk(bountyInfo,  { include: db.models.Bounty.Company });
+	} else {
+		return await db.models.Bounty.findOne({ where: { userId: bountyInfo.posterId, companyId: guildId, slotNumber: bountyInfo.slotNumber, state: "open" }, include: db.models.Bounty.Company });
+	}
+}
+
+/**
  * @param {Bounty} bounty
+ * @param {Guild} guild
  * @param {Company} company
```
ntseng, line 36: `company` isn't in the parameters of this function anymore.

source/logic/bounties.js
```diff
@@ -49,7 +84,9 @@ async function addCompleters(guild, bounty, completerIds) {
 		bounty,
 		allCompleters,
 		poster,
-		company
+		company,
```
ntseng, line 87: `allCompleters`, `poster`, and `company` are all just being queried at the end of this function without anything else depending on them. I think it'd be better to fetch them via their own endpoints instead of coupling them with this one.

Specific Thoughts:
- `allCompleters`: I think it'd be reasonable to have this file also handle Completions since they depend on Bounties
- `poster`: are we coding defensively enough that we aren't assuming a bounty existing implies both the `poster`'s `Hunter` and `User`? If so, we need more error handling, otherwise this would want a separate endpoint from `findOrCreateBountyHunter()` to skip the "orCreate" contingency.

SillySalamndr: This is another artifact of the Java mindset that I brought in at the beginning of this effort, and I've already said my peace on that. I'm noticing that changes like this are starting to gum up the flow of getting this resolved, so making a new issue for this might help get things going (and even let you pick this up if I can't get to it fast enough).

ntseng:
Bug to be aware of (but not needed for this PR): #341

ntseng:
Updated PR title to include a description of the work. The commit title will also need to be updated when it gets merged.

SillySalamndr:
Merge should be done. Please do a once-over to make sure I haven't missed anything, pleases.

SillySalamndr:
Also, because it got ate: I captured the future effort for discussing (and maybe fixing) error/exception handling in https://github.com/Imaginary-Horizons-Productions/BountyBot/issues/378.

source/buttons/bbaddcompleters.js
```diff
@@ -37,61 +38,52 @@ async function updateBoardPosting(bounty, company, poster, newCompleterIds, comp
...
+					new UserSelectMenuBuilder().setCustomId(SKIP_INTERACTION_HANDLING)
+						.setPlaceholder("Select bounty hunters...")
+						.setMaxValues(5)
+				)
+			],
+			flags: [MessageFlags.Ephemeral],
+			withResponse: true
+		}).then(response => response.resource.message.awaitMessageComponent({ time: timeConversion(2, "m", "ms"), componentType: ComponentType.UserSelect })).then(async collectedInteraction => {
+			try {
+			let { bounty: returnedBounty, allCompleters, poster, company, validatedCompleterIds } = await logicLayer.bounties.addCompleters(collectedInteraction.guild, bounty, validatedCompleterIds);
```
ntseng, line 64: Using the Bounty Board Add Completers Button fails with the following error:
image

source/commands/bounty/addcompleters.js
```diff
@@ -1,6 +1,7 @@
 const { CommandInteraction, userMention, bold, MessageFlags, Guild } = require("discord.js");
 const { Sequelize } = require("sequelize");
 const { extractUserIdsFromMentions, listifyEN, commandMention, congratulationBuilder } = require("../../util/textUtil");
+const { addCompleters, findBounty } = require("../../logic/bounties.js");
```
ntseng, line 4: Bad merge resolution?

ntseng (changes requested):
Found a critical path failure and some logic layer instance mishandling (merge resolution error?).

source/commands/bounty/addcompleters.js
SillySalamndr: So, in the merge resolution, for some unknown reason, the logic bundle just...didn't exist. To try to resolve this at least locally, I left in this line. I'll try to update this later when `main` is in a better state, but likely not before we meet today.

source/context_menus/Give_Bounty_Credit.js
```diff
@@ -1,6 +1,7 @@
 const { InteractionContextType, PermissionFlagsBits, ModalBuilder, ActionRowBuilder, TextInputBuilder, TextInputStyle, userMention, bold, MessageFlags, Guild } = require('discord.js');
 const { UserContextMenuWrapper } = require('../classes');
 const { SKIP_INTERACTION_HANDLING } = require('../constants');
+const { addCompleters, findBounty } = require('../logic/bounties.js');
```
SillySalamndr, line 4: This spot _does_ have the logic layer, though, so I should make a point to remove this.

source/buttons/bbaddcompleters.js
SillySalamndr: For some other unknown reason, this comment isn't being put in context of the files changed. I'll be trying to get this done before our meeting today, but it might take me some time to find it. (I'm also noticing that the indentation got ruined here...I'm not sure I like the web merge resolver, and might be trying to find a different solution.)

SillySalamndr: Ah, here it is.

SillySalamndr: Yeah, this got messed up in the merge.

source/commands/bounty/addcompleters.js
SillySalamndr: Okay, I'm starting to see some evidence elsewhere that speaks to the contrary. Let me see if I can get a fix in ~20 minutes

SillySalamndr: Okay, it'll be good with the next commit

source/logic/bounties.js
```diff
@@ -16,6 +18,18 @@ function setDB(database) {
...
+ * @param {{slotNumber: number, posterId: string, guildId: string} | string} bountyInfo
+ * @returns {Bounty?} the bounty, if it exists
+ */
+async function findBounty(bountyInfo) {
+	if (typeof bountyInfo === 'string') {
+		return await db.models.Bounty.findByPk(bountyInfo,  { include: db.models.Bounty.Company });
+	} else {
+		return await db.models.Bounty.findOne({ where: { userId: bountyInfo.posterId, companyId: guildId, slotNumber: bountyInfo.slotNumber, state: "open" }, include: db.models.Bounty.Company });
+	}
+}
```
ntseng, line 31:
```suggestion
/**
 * @param {{slotNumber: number, posterId: string, guildId: string} | string} bountyInfo
 */
function findBounty(bountyInfo) {
	if (typeof bountyInfo === 'string') {
		return db.models.Bounty.findByPk(bountyInfo,  { include: db.models.Bounty.Company });
	} else {
		return db.models.Bounty.findOne({ where: { userId: bountyInfo.posterId, companyId: guildId, slotNumber: bountyInfo.slotNumber, state: "open" }, include: db.models.Bounty.Company });
	}
}
```
In this function, I prefer to omit the `async-await` because the only thing being awaited is the final return, which practically doesn't change things. Further, it's usually better to allow VS Code detect the return type of the function, as seeing something out of place can help find bugs, and it doesn't have potential for outdated documentation desyncs.

Minimally, the return type is incorrect: should be `Promise<Bounty?>` instead of `Bounty?`.

PR #327, opened 2025-02-19
#321: move embed building out of `raiseToast()`

### LINKED ISSUE #321: Move embed building in toast flows out of `raiseToast()`
Building a toast embed is a frontend responsibility that is currently in the logic layer because that usage is coupled with server goal tracking. See comments for #306 for more details.

ntseng:
Summary
-------
- move embed building out of `raiseToast()`
- consolidated the generation of the following message payload components
   - toast embed
   - goal completion embed
   - toast seconding action row

Local Tests Performed
---------------------
- [x] toast by command still resolves
- [x] toast by context menu still resolves

Issue
-----
Closes #321

### DIFF
--- a/source/models/companies/Goal.js
+++ b/source/models/companies/Goal.js
@@ -1,4 +1,6 @@
+const { EmbedBuilder, userMention } = require('discord.js');
 const { Model, Sequelize, DataTypes } = require('sequelize');
+const { congratulationBuilder, listifyEN } = require('../../util/textUtil');
 
 /** A Goal for which all bounty hunters in a company contribute to */
 class Goal extends Model {
@@ -6,6 +8,15 @@ class Goal extends Model {
 		models.Goal.Contributions = models.Goal.hasMany(models.Contribution, {
 			foreignKey: "goalId"
 		})
+	}
+
+	/** @param {string[]} contributorIds */
+	static generateCompletionEmbed(contributorIds) {
+		return new EmbedBuilder().setColor("e5b271")
+			.setTitle("Server Goal Completed")
+			.setThumbnail("https://cdn.discordapp.com/attachments/673600843630510123/1309260766318166117/trophy-cup.png?ex=6740ef9b&is=673f9e1b&hm=218e19ede07dcf85a75ecfb3dde26f28adfe96eb7b91e89de11b650f5c598966&")
+			.setDescription(`${congratulationBuilder()}, the Server Goal was completed! Contributors have double chance to find items on their next bounty completion.`)
+			.addFields({ name: "Contributors", value: listifyEN(contributorIds.map(id => userMention(id))) })
 	}
 }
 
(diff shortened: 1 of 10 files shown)


source/models/toasts/Toast.js
```diff
@@ -20,6 +20,30 @@ class Toast extends Model {
...
+	 * @param {string[]} recipientIds
+	 * @param {GuildMember} senderMember
+	 */
+	static generateEmbed(thumbnailURL, toastText, recipientIds, senderMember) {
+		return new EmbedBuilder().setColor("e5b271")
+			.setThumbnail(thumbnailURL ?? 'https://cdn.discordapp.com/attachments/545684759276421120/751876927723143178/glass-celebration.png')
+			.setTitle(toastText)
+			.setDescription(`A toast to ${listifyEN(recipientIds.map(id => userMention(id)))}!`)
+			.setFooter({ text: senderMember.displayName, iconURL: senderMember.user.avatarURL() });
+	}
```
SillySalamndr, line 35: This feels very weird as a static function, since the embed we're generating is about a toast object. Why is this not a function on the model instance? You would potentially drop some arguments needed to be passed, right?

source/models/toasts/Toast.js
```diff
@@ -20,6 +20,30 @@ class Toast extends Model {
...
+
+	/** @param {string} toastId */
+	static generateSecondingActionRow(toastId) {
+		return new ActionRowBuilder().addComponents(
+			new ButtonBuilder().setCustomId(`secondtoast${SAFE_DELIMITER}${toastId}`)
+				.setLabel("Hear, hear!")
+				.setEmoji("🥂")
+				.setStyle(ButtonStyle.Primary)
+		)
+	}
```
SillySalamndr, line 45: Similar here as with the generateEmbed function. We'd drop the need for any arguments here.

source/models/companies/Goal.js
```diff
@@ -7,6 +9,15 @@ class Goal extends Model {
...
 	}
+
+	/** @param {string[]} contributorIds */
+	static generateCompletionEmbed(contributorIds) {
+		return new EmbedBuilder().setColor("e5b271")
+			.setTitle("Server Goal Completed")
+			.setThumbnail("https://cdn.discordapp.com/attachments/673600843630510123/1309260766318166117/trophy-cup.png?ex=6740ef9b&is=673f9e1b&hm=218e19ede07dcf85a75ecfb3dde26f28adfe96eb7b91e89de11b650f5c598966&")
+			.setDescription(`${congratulationBuilder()}, the Server Goal was completed! Contributors have double chance to find items on their next bounty completion.`)
+			.addFields({ name: "Contributors", value: listifyEN(contributorIds.map(id => userMention(id))) })
+	}
```
SillySalamndr, line 20: Same idea here as with the Toast comments. It seems like the only data used here is the co tributor IDs, and it would be more clear if we could fetch them directly off of the goal model, instead.

SillySalamndr (approved):
If it is reasonable to have the embed building use functions on the concrete objects instead of as static methods, I think we will gain a little bit of clarity overall. A change like this might be more work than is worth it in this moment, and the current effort is a step in the right direction regardless. If you agree with this, then please make an issue(s) to address this in the future instead of needing to do so now.

ntseng:
I actually looked into non-static functions and as of today, neither the `Toast` nor the `Goal` instance are kept in the frontend for the generator functions. I agree the non-static versions are easier to read, but feel it's low enough priority to wait until a refactor that surfaces the instance objects happens.

Issue #310, opened 2025-02-07
Migrate `buildModStatsEmbed()` from utils to logic layer


ntseng:
Moving this function to the logic layer (hunters branch) will allow `/moderation user-report` to not need database as an argument.

SillySalamndr:
Putting the new modStatsEmbed on the Hunter will allow for the logic layer to do the data fetching and the model to do the representation management

Issue #309, opened 2025-02-07
Migrate `buildSeasonalScoreboardEmbed()` and `buildOverallScoreboardEmbed()` from utils to logic layer


ntseng:
Moving these two functions to the logic layer (companies branch) will allow `updateScoreboard()` to not need database as an argument.

SillySalamndr:
Putting the new scoreboard embed generators on the Company will allow for the logic layer to do the data fetching and the model to do the representation management

ntseng:
Duplicate of #346

Issue #308, opened 2025-02-07
Migrate `showcaseBounty()` from utils to logic and frontend layers


ntseng:
Validations and interaction acknowledgement should go in frontend, logic in logic.

PR #329, opened 2025-02-19
create logic file for Seasons

ntseng:
Summary
-------
- create logic file for Seasons
- pass reference to single database connection around command layer
- create setItemLogic
- rename "logicBundle" to "logicBlob" as both refer to the same shape
- fix syntax error in setLogic iterators (treating iterator of for..in as object instead of string)
- add logicLayer boiler plate to blueprints
   - left direct imports from logic layer in util files, as out of scope (already scheduled as #308, #309, and #310)
- add deduplication to Completers on bounty embed (for evergreen bounties)

Local Tests Performed
---------------------
- [x] used User Stats context menu flow to ensure fetching from database still works
- [x] used `/bounty list (bot mention)` to ensure evergreen bounty data can be fetched from database from subcommands

### DIFF
--- a/source/models/bounties/Bounty.js
+++ b/source/models/bounties/Bounty.js
@@ -2,7 +2,7 @@
 const { Model, Sequelize, DataTypes } = require('sequelize');
 const { Company } = require('../companies/Company');
 const { SAFE_DELIMITER, MAX_MESSAGE_CONTENT_LENGTH } = require('../../constants');
-const { timeConversion, commandMention } = require('../../util/textUtil');
+const { timeConversion, commandMention, listifyEN } = require('../../util/textUtil');
 
 /** Bounties are user created objectives for other server members to complete */
 class Bounty extends Model {
@@ -51,7 +51,8 @@ class Bounty extends Model {
 				embed.setAuthor({ name: `${author.displayName}'s #${this.slotNumber} Bounty`, iconURL: author.user.displayAvatarURL() });
 			}
 			if (completions.length > 0) {
-				fields.push({ name: "Completers", value: `<@${completions.map(reciept => reciept.userId).join(">, <@")}>` });
+				const uniqueCompleters = new Set(completions.map(reciept => reciept.userId));
+				fields.push({ name: "Completers", value: listifyEN([...uniqueCompleters].map(id => userMention(id))) });
 			}
 
 			if (fields.length > 0) {


source/bot.js
```diff
@@ -223,7 +224,7 @@ client.on(Events.GuildDelete, async guild => {
 	db.models.Completion.destroy({ where: { companyId: guild.id } });
 
 	db.models.Participation.destroy({ where: { companyId: guild.id } });
-	db.models.Season.destroy({ where: { companyId: guild.id } });
+	deleteCompanySeasons(guild.id);
```
SillySalamndr, line 227: This shouldn't work. The instance that you're using here is not the instance that is being supplied with the database during the setup steps since it is a separate import. Since this is a one off call, it is probably cleanest to replace this with `logicBlob['seasons'].deleteCompanySeasons(guild.id);`.

source/buttons/bbtakedown.js
```diff
@@ -33,7 +34,7 @@ module.exports = new ButtonWrapper(mainId, 3000,
 
 				findOneHunter(interaction.user.id, interaction.guild.id).then(async hunter => {
 					hunter.decrement("xp");
-					const [season] = await database.models.Season.findOrCreate({ where: { companyId: interaction.guildId, isCurrentSeason: true } });
+					const [season] = await findOrCreateCurrentSeason(interaction.guild.id);
```
SillySalamndr, line 37: This instance of the function isn't supplied with the database.

source/buttons/secondtoast.js
```diff
@@ -41,7 +42,7 @@ module.exports = new ButtonWrapper(mainId, 3000,
 				recipientIds.push(reciept.recipientId);
 			}
 		});
-		const [season] = await database.models.Season.findOrCreate({ where: { companyId: interaction.guildId, isCurrentSeason: true } });
+		const [season] = await findOrCreateCurrentSeason(interaction.guild.id);
```
SillySalamndr, line 45: This instance of the function isn't supplied with the database.

source/commands/bounty/post.js
```diff
@@ -178,7 +179,7 @@ async function executeSubcommand(interaction, database, runMode, ...[posterId, h
 				return;
 			}
 
-			const [season] = await database.models.Season.findOrCreate({ where: { companyId: modalSubmission.guildId, isCurrentSeason: true } });
+			const [season] = await findOrCreateCurrentSeason(modalSubmission.guild.id);
```
SillySalamndr, line 182: This instance of the function isn't supplied with the database.

source/commands/bounty/takedown.js
```diff
@@ -43,7 +44,7 @@ async function executeSubcommand(interaction, database, runMode, ...[posterId])
 
 			findOneHunter(interaction.user.id, interaction.guild.id).then(async hunter => {
 				hunter.decrement("xp");
-				const [season] = await database.models.Season.findOrCreate({ where: { companyId: interaction.guildId, isCurrentSeason: true } });
+				const [season] = await findOrCreateCurrentSeason(interaction.guild.id);
```
SillySalamndr, line 47: This instance of the function isn't supplied with the database.

source/commands/evergreen/complete.js
```diff
@@ -22,7 +23,7 @@ async function executeSubcommand(interaction, database, runMode, ...args) {
 	}
 
 	const company = await database.models.Company.findByPk(interaction.guildId);
-	const [season] = await database.models.Season.findOrCreate({ where: { companyId: interaction.guildId, isCurrentSeason: true } });
+	const [season] = await findOrCreateCurrentSeason(interaction.guildId);
```
SillySalamndr, line 26: This instance of the function isn't supplied with the database.

source/commands/moderation/seasondisqualify.js
```diff
@@ -12,7 +13,7 @@ const { findOrCreateCompany } = require("../../logic/companies");
 async function executeSubcommand(interaction, database, runMode, ...args) {
 	const member = interaction.options.getMember("bounty-hunter");
 	await findOrCreateCompany(interaction.guild.id);
-	const [season] = await database.models.Season.findOrCreate({ where: { companyId: interaction.guildId, isCurrentSeason: true } });
+	const [season] = await findOrCreateCurrentSeason(interaction.guildId);
```
SillySalamndr, line 16: This instance of the function isn't supplied with the database.

source/commands/moderation/takedown.js
```diff
@@ -49,7 +50,7 @@ async function executeSubcommand(interaction, database, runMode, ...args) {
 
 			findOneHunter(posterId, interaction.guild.id).then(async poster => {
 				poster.decrement("xp");
-				const [season] = await database.models.Season.findOrCreate({ where: { companyId: interaction.guildId, isCurrentSeason: true } });
+				const [season] = await findOrCreateCurrentSeason(interaction.guildId);
```
SillySalamndr, line 53: This instance of the function isn't supplied with the database.

source/commands/moderation/xppenalty.js
```diff
@@ -19,7 +20,7 @@ async function executeSubcommand(interaction, database, runMode, ...args) {
 	const penaltyValue = Math.abs(interaction.options.getInteger("penalty"));
 	hunter.decrement({ xp: penaltyValue });
 	hunter.increment({ penaltyCount: 1, penaltyPointTotal: penaltyValue });
-	const [season] = await database.models.Season.findOrCreate({ where: { companyId: interaction.guildId, isCurrentSeason: true } });
+	const [season] = await findOrCreateCurrentSeason(interaction.guildId);
```
SillySalamndr, line 23: This instance of the function isn't supplied with the database.

source/commands/season-end.js
```diff
@@ -13,12 +14,12 @@ module.exports = new CommandWrapper(mainId, "Start a new season for this server,
...
 
 		buildCompanyStatsEmbed(interaction.guild, database).then(async embed => {
-			const seasonBeforeEndingSeason = await database.models.Season.findOne({ where: { companyId: interaction.guildId, isPreviousSeason: true } });
+			const seasonBeforeEndingSeason = await findOneSeason(interaction.guildId, "previous");
 			if (seasonBeforeEndingSeason) {
 				seasonBeforeEndingSeason.isPreviousSeason = false;
 				seasonBeforeEndingSeason.save();
 			}
-			const endingSeason = await database.models.Season.findOne({ where: { companyId: interaction.guildId, isCurrentSeason: true } });
+			const endingSeason = await findOneSeason(interaction.guildId, "current");
```
SillySalamndr, line 22: This instance of the function isn't supplied with the database.

source/commands/season-end.js
```diff
@@ -41,7 +42,7 @@ module.exports = new CommandWrapper(mainId, "Start a new season for this server,
 				endingSeason.isPreviousSeason = true;
 				endingSeason.save();
 			}
-			await database.models.Season.create({ companyId: interaction.guildId });
+			await createSeason(interaction.guildId);
```
SillySalamndr, line 45: This instance of the function isn't supplied with the database.

SillySalamndr:
It feels like I might've missed something somewhere along the line, but it truly appears like all of the new logic file implementations are missing the part where they're being passed into the command files, which means that none of the command files have access to an instance of the logic files with the database association.

ntseng:
@SillySalamndr ready for re-review!

source/buttons/_button_blueprint.js
```diff
@@ -1,9 +1,14 @@
 const { ButtonWrapper } = require('../classes');
 
+/** @type {typeof import("../logic")} */
+let logicLayer;
+
```
SillySalamndr, line 5: As much as I'd personally prefer having separate variables for the specific logic files as a means to signal what is being done/used in the command-style file, this does resolve all possible import situations/issues, and it isn't bad for performance. This is almost certainly better than accidentally missing the proper import/assignment in the logic link.

source/classes/InteractionWrapper.js
```diff
@@ -18,6 +18,12 @@ class InteractionWrapper {
 		this.execute = executeFunction;
 	}
 
+	/** @param {(logicBlob: typeof import("../logic")) => void} setLogicFunction */
+	setLogicLinker(setLogicFunction) {
+		this.setLogic = setLogicFunction;
+		return this;
+	}
+
```
SillySalamndr, line 26: Yeah, builder pattern works well here.

source/util/scoreUtil.js
SillySalamndr, line 1: This file doesn't seem to have gotten the same treatment for updating logic file associations that you've applied elsewhere, in case this was missed somehow.

source/util/embedUtil.js
SillySalamndr, line 1: This file doesn't seem to have gotten the same treatment for updating logic file associations that you've applied elsewhere, in case this was missed somehow.

source/logic/toasts.js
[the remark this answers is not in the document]
SillySalamndr: I'm not going to go and add this comment to all the logic files that use other logic files for some data loading/processing, but it applies to the rest, too.

source/items/profile-colorizer-purple.js
```diff
@@ -1,16 +1,20 @@
...
 const color = "Purple";
 const itemName = `${color} Profile Colorizer`;
 module.exports = new Item(itemName, `Changes the color of your stats profile embed to ${color.toLowerCase()}`, 3000,
 	/** Sets the user's Hunter profile to Colors.Purple in the used guild */
 	async (interaction, database) => {
-		findOneHunter(interaction.user.id, interaction.guild.id).then(hunter => {
+		logicLayer.hunters.findOneHunter(interaction.user.id, interaction.guild.id).then(hunter => {
 			hunter.profileColor = color;
 			hunter.save();
 		})
```
SillySalamndr, line 15: So that the entire database transaction happens in the logic layer, I'd expect that the function used here would be a "setHunterColor" that passes the hunter-identifying information and the color to have all that effort happen at the logic layer.

source/items/profile-colorizer-blue.js
SillySalamndr, line 1: I'm starting to notice that the code in each of these is incredibly repetitive, with only one "real" parameter changing between them. Maybe a factory pattern would be a way to reduce the repetition here? (Probably not work for this specific task.)

source/util/scoreUtil.js
ntseng: Yep, should still be a problem here. The util files have their own issues (#333 and others).

source/util/embedUtil.js
ntseng: See reply on scoreUtil comment.

source/logic/toasts.js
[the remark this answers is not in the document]
SillySalamndr: There's likely two ways forward with this:

1. Each logic file has a `setLogic` similar to `setDB` to receive "real" links to the other logic files.
2. Logic files being largely independent of other logic files is likely a good boon to maintain. We could instead have logic files that use another logic file create their own instance of said logic file and run `setDB` on the specific logic files that they use, additionally using the database link they receive to set the DB properly on children. (Risk: recursive loop)

SillySalamndr (commented):
This is...close. There's a couple of concerns, but half of those should not be addressed in this pull request, most likely, and it might be sensible to address the other more directly associated changes in separate issues/pull requests due to how massive this pull request has gotten.

source/logic/toasts.js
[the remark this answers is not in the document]
ntseng: I suspect, but haven't confirmed, using other logic file endpoints is an indicator of stolen responsibility. I think we should revisit towards the end of the overhaul in case the refactors just make the direct imports go away on their own.

source/items/profile-colorizer-purple.js
ntseng: Good idea: #336

source/items/profile-colorizer-blue.js
ntseng: This feels like it's on the verge of overengineering: squeezing a factory in between item file reading and item dictionary population feels like a big interface change.

source/logic/toasts.js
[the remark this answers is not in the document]
ntseng: Did a quick survey and I don't think there's a need to distribute layer references among the different logic files. All the direct imports fall into the following two categories:
1. `raiseToast()` and `progressGoal()` have too much responsibility and should be split up
2. there's a minor loss of convenience on not having simplified queries (`findOneSeason()`) and some commonly grouped queries (`findOrCreateBountyHunter()`), but those functions are simple enough that it's probably better to inline them than enforce them as top-level concepts

Category 1 is too big for the scope of this PR, and Category 2 should just go away on its own as we do cleanup around the codebase.

source/items/profile-colorizer-blue.js
ntseng: Interesting difference in perspective here: I would have classed our number of profile colorizers as "kinda big" (as opposed to "truly massive"); maybe that's bias from PotL where "instance edge case = content".

Personally, my biggest complaint against code duplication is "it makes it easy to forget to apply bugfixes and optimizations to all uses of a functionality". I think another reason I'm deprioritizing this is because the enumerated list of profile colorizers in itemDictionary mitigates that risk pretty well. If we found another family of items that could benefit from a factory (hopefully more than these) then I'd be more onboard with investing in the factory tech.

source/logic/toasts.js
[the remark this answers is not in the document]
ntseng: Mostly empty reminder issue made: #337

PR #339, opened 2025-02-24
#310: move `buildModStatsEmbed()` to `Hunter.modStatsEmbed()`

### LINKED ISSUE #310: Migrate `buildModStatsEmbed()` from utils to logic layer

ntseng:
Summary
-------
- move `buildModStatsEmbed()` to `Hunter.modStatsEmbed()`
- move `buildCompanyStatsEmbed()` to `Company.statsEmbed()`

This is Ugly but is Scheduled to Be Refactored Later
------------
- `company` removed from `updateScoreboard()` parameters to avoid circular dependency with Company (which needs `ihpAuthorPayload` among others)
- added `logicLayer` as parameter for `buildSeasonalScoreboardEmbed()` and `buildOverallScoreboardEmbed()`

Local Tests Performed
---------------------
- [x] used context menu flow to check user stats
- [x] used context menu flow to check company stats

Issue
-----
Closes #310

### DIFF
--- a/source/util/embedUtil.js
+++ b/source/util/embedUtil.js
@@ -1,13 +1,8 @@
 const fs = require("fs");
 const { EmbedBuilder, Guild, Colors } = require("discord.js");
 const { Sequelize } = require("sequelize");
-const { Hunter } = require("../models/users/Hunter");
-const { Company } = require("../models/companies/Company");
-const { COMPANY_XP_COEFFICIENT, MAX_EMBED_DESCRIPTION_LENGTH } = require("../constants");
+const { MAX_EMBED_DESCRIPTION_LENGTH } = require("../constants");
 const { generateTextBar } = require("./textUtil");
-const { findLatestGoalProgress } = require("../logic/goals");
-const { findOrCreateCompany } = require("../logic/companies");
-const { findOrCreateCurrentSeason, findOneSeason } = require("../logic/seasons");
 
 const discordIconURL = "https://cdn.discordapp.com/attachments/618523876187570187/1110265047516721333/discord-mark-blue.png";
 const bountyBotIcon = "https://cdn.discordapp.com/attachments/618523876187570187/1138968614364528791/BountyBotIcon.jpg";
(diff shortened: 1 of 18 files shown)


source/util/embedUtil.js
```diff
@@ -260,27 +188,25 @@ async function buildVersionEmbed() {
...
  */
-function updateScoreboard(company, guild, database) {
+async function updateScoreboard(guild, database, logicLayer) {
+	const [company] = await logicLayer.companies.findOrCreateCompany(guild.id);
 	if (company.scoreboardChannelId && company.scoreboardMessageId) {
 		guild.channels.fetch(company.scoreboardChannelId).then(scoreboard => {
 			return scoreboard.messages.fetch(company.scoreboardMessageId);
 		}).then(async scoreboardMessage => {
-			scoreboardMessage.edit({ embeds: [company.scoreboardIsSeasonal ? await buildSeasonalScoreboardEmbed(guild, database) : await buildOverallScoreboardEmbed(guild, database)] });
+			scoreboardMessage.edit({ embeds: [company.scoreboardIsSeasonal ? await buildSeasonalScoreboardEmbed(guild, database, logicLayer) : await buildOverallScoreboardEmbed(guild, database, logicLayer)] });
```
SillySalamndr, line 200: The pass-along of the logic layer feels weird here, especially since one of these locations (the overall scoreboard) is only using the logic layer to find or make a Company, which we're already doing here. There seems to be an opportunity to have the company passed in to this function instead of using the logic layer directly within this context, but this amount of extra work is certainly out of scope. I'd also expect that these functions are going to be moved around a bit, which would also complicate things.

ntseng: Yeah, I'm not a fan either (even gave it dishonorable mention in the PR description), but scoreboard generator refactors deserve their own PR.

SillySalamndr (approved):
Overall LGTM. I created an issue to potentially address the logic layer passing to the util prompt generation, but it overall isn't a thing that I'd address in this issue if at all. There's the potential for performance issues with hitting the database multiple times for the same data, but that is a pretty low risk, overall, and this kind of issue might be addressed inadvertently in a different task.

PR #340, opened 2025-02-24
create `logicLayer.companies.findCompanyByPK()`

ntseng:
Summary
-------
- create `logicLayer.companies.findCompanyByPK()`

Local Tests Performed
---------------------

### DIFF
--- a/source/commands/bounty/post.js
+++ b/source/commands/bounty/post.js
@@ -50,8 +50,8 @@ async function executeSubcommand(interaction, database, runMode, ...[logicLayer,
 	}).then(response => response.resource.message.awaitMessageComponent({ time: 120000, componentType: ComponentType.StringSelect })).then(async collectedInteraction => {
 		const [slotNumber] = collectedInteraction.values;
 		// Check user actually has slot
-		const company = await database.models.Company.findByPk(interaction.guildId);
-		const hunter = await logicLayer.hunters.findOneHunter(interaction.user.id, interaction.guild.id);
+		const company = await logicLayer.companies.findCompanyByPK(interaction.guild.id);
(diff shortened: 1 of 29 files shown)


source/commands/bounty/post.js
```diff
@@ -50,8 +50,8 @@ async function executeSubcommand(interaction, database, runMode, ...[logicLayer,
 	}).then(response => response.resource.message.awaitMessageComponent({ time: 120000, componentType: ComponentType.StringSelect })).then(async collectedInteraction => {
 		const [slotNumber] = collectedInteraction.values;
 		// Check user actually has slot
-		const company = await database.models.Company.findByPk(interaction.guildId);
-		const hunter = await logicLayer.hunters.findOneHunter(interaction.user.id, interaction.guild.id);
+		const company = await logicLayer.companies.findCompanyByPK(interaction.guild.id);
+		const hunter = await logicLayer.hunters.findOneHunter(interaction.user.id, interaction.guildId);
```
SillySalamndr, line 54: `interaction.guild.id` vs. `interaction.guildId` here? Which one being used where is flipped here, and I can't tell for what reason. Was this intentional? Can we just use one or the other?

SillySalamndr (commented):
Overall LGTM. One oddity that's captured in a comment, but I don't believe that it is especially severe, other than some consistency.

source/commands/bounty/post.js
SillySalamndr: Reasoning: Micro-optimization
1st `guild.id` crashes early with "could not find guild" as a more useful error message
2nd `guildId` is faster access and we now know that we have a guild

SillySalamndr (approved):

PR #343, opened 2025-02-25
create rank logic branch

ntseng:
Summary
-------
- create rank logic branch
- fix crash on `/create-default rank-roles` when rank exists at a default variance threshold

Future Work
-----
- Optimize `/create-default rank-roles` acknowledge within 3 seconds (to avoid needing to `deferReply()`)

Local Tests Performed
---------------------
- [x] `/create-default rank-roles` still writes to the database

### DIFF
--- a/source/logic/ranks.js
+++ b/source/logic/ranks.js
@@ -1,0 +1,71 @@
+const { Sequelize } = require("sequelize");
+
+/** @type {Sequelize} */
+let db;
+
+/** *Set the database pointer for the Rank logic file*
+ * @param {Sequelize} database
+ */
+function setDB(database) {
+	db = database;
+}
+
+const DEFAULT_VARIANCE_THRESHOLDS = [2.5, 1, 0, -3];
+const DEFAULT_RANKMOJI = ["🏆", "🥇", "🥈", "🥉"];
+
+/** *Creates Ranks for a Company with the default emoji and variance thresholds*
+ * @param {string} companyId
+ * @param {string[]} roleIds
+ */
+function createDefaultRanks(companyId, roleIds) {
+	return db.models.Rank.bulkCreate(roleIds.map((roleId, index) => ({
+		companyId,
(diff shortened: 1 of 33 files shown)


source/logic/ranks.js
```diff
@@ -0,0 +1,71 @@
...
+ * @param {"ascending" | "descending"} sortOrder the order to sort the ranks by variance threshold
+ */
+function findAllRanks(companyId, sortOrder) {
+	switch (sortOrder) {
+		case "ascending":
+			return db.models.Rank.findAll({ where: { companyId }, order: [["varianceThreshold", "ASC"]] });
+		case "descending":
+			return db.models.Rank.findAll({ where: { companyId }, order: [["varianceThreshold", "DESC"]] });
+	}
+}
```
SillySalamndr, line 55: This means of changing the order is probably fine, but I wonder if the use of this function is slightly clearer if we don't pass a sort order and always return the same order, with the intent to do `Array.prototype.reverse()` in places we want the opposite ordering scheme. Likely not worth changing at this point, or maybe even ever, with it already being implemented as-is.

SillySalamndr (approved):
Overall LGTM. I'm not sure if being able to pass the role ordering is that much of a performance uplift over the clarity of always returning one ordering and doing an `Array.prototype.reverse()` in the cases where we want the other order, but the current implementation looks good enough regardless that there's next to no _real_ reason to change it.

source/logic/ranks.js
ntseng: Since ranks are "very small n", I have a very weak opinion on what we implement for this specific case, but I do prefer this pattern (have used it elsewhere already).

The reasoning being:
- The database should be more efficient at manipulating entity order
- In order to avoid exponential endpoint variations, I'm hoping to keep some simpler query specifications as parameters (ie `findAllRanks(guild.id, "ascending")` vs `findAllRanksAscneding(guild.id)` scales better if we later want to add a filter of some kind directly into the endpoint)

PR #344, opened 2025-02-25
create `logicLayer.bounties.findOpenBounties()`

ntseng:
Summary
-------
- create `logicLayer.bounties.findOpenBounties()`
- create `logicLayer.bounties.findEvergreenBounties()`

Local Tests Performed
---------------------
- [x] `/evergreen post`, `/evergreen swap`, `/evergreen edit`, and `/evergreen take-down` still work

### DIFF
--- a/source/commands/evergreen/takedown.js
+++ b/source/commands/evergreen/takedown.js
@@ -11,7 +11,7 @@ const { bountiesToSelectOptions } = require("../../util/messageComponentUtil");
  * @param {[typeof import("../../logic")]} args
  */
 async function executeSubcommand(interaction, database, runMode, ...[logicLayer]) {
-	const openBounties = await database.models.Bounty.findAll({ where: { companyId: interaction.guildId, userId: interaction.client.user.id, state: "open" }, order: [["slotNumber", "ASC"]] });
+	const openBounties = await logicLayer.bounties.findEvergreenBounties(interaction.guild.id);
 	interaction.reply({
 		content: `If you'd like to change the title, description, or image of an evergreen bounty, you can use ${commandMention("evergreen edit")} instead.`,
 		components: [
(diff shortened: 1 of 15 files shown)


source/commands/evergreen/takedown.js
```diff
@@ -26,14 +26,13 @@ async function executeSubcommand(interaction, database, runMode, ...[logicLayer]
 		withResponse: true
 	}).then(response => response.resource.message.awaitMessageComponent({ time: 120000, componentType: ComponentType.StringSelect })).then(async collectedInteraction => {
 		const [bountyId] = collectedInteraction.values;
-		const bounty = await database.models.Bounty.findByPk(bountyId, { include: database.models.Bounty.Company });
+		const [bounty] = openBounties.splice(openBounties.findIndex(bounty => bounty.id === bountyId), 1);
```
SillySalamndr, line 29: Using `splice` here feels weird, to be honest. I think `filter` might be a better solution, since you would be able to roll the `findIndex` functionality into the first function and not need to splice exactly 1 result. Since the `findIndex` function searches by ID, it is also reasonable that we only get 1 result out of this method.

SillySalamndr (commented):
I think there's only 1 thing that could be changed here, and it is pretty minor and isn't a potential bug.

source/commands/evergreen/takedown.js
ntseng: As part of "one of frontend responsibilities is managing the resource store", the side-effect of removing the bounty element from `openBounties` is utilized later on.

SillySalamndr: Again, I don't think this pattern is bad, per se, I just think that `filter` does the job here better than `splice` and is more direct about the purpose of the code being executed here. (We're filtering down the original list to just the element with the given ID.)

ntseng: Doesn't `filter` not give you reference to the removed stuff? We need that as well to tell the database to `destroy()`.

SillySalamndr:
```suggestion
		const [bounty] = openBounties.filter(bounty => bounty.id === bountyId);
```
As much as this suggestion doesn't _guarantee_ an array with a single element, the database should be enforcing that there's only one bounty with each ID.

ntseng: Oh, huh. I'd probably use `.find()` for that specific case since it ends on match. Even then though, `.filter()` doesn't change the original array, so the later `openBounties` checks would need to account for that.

To some extent, this will get solved automatically when we make the `deleteBounty()` endpoint, as then we won't need references to both the bounty to be removed and the remaining bounties in the frontend.

SillySalamndr: Yeah, `find` would also do the trick in the same way, but if we ever got paranoid of duplicate IDs, `filter` would give us an opportunity to get the duplicates. (We likely don't need to and shouldn't need to worry about this, though.)

I'm not a fan of needing to depend on the side effect of `splice` here, though. I definitely didn't catch that, since I'm not used to code needing that side effect, and if they do need that side effect, they overwrite the original variable in the process to make that intent clear. This also makes a bit of the later code confusing, since `openBounties` _should_ only be one element at that point (and we should only have 1 bounty per bounty ID). If that's the case, why not just use `bounty` in those cases, instead?

ntseng: I guess the clarity issues are real. `openBounties` at that point is all evergreen bounties except the one about to be deleted.

SillySalamndr: Ah, I forgot entirely about that feature of `splice`, which _does_ make it reasonable to use here. I'm still not enthusiastic about the side effects usage, but it _is_ useful for the critical path of the code.

SillySalamndr (approved):

PR #347, opened 2025-03-02
fix crash on add completers button in archived thread

ntseng:
Summary
-------
- fix crash on add completers button in archived thread
- add missing type annotation imports

Local Tests Performed
---------------------
- [x] bot doesn't crash if adding completers to an archived bounty thread via the thread button


SillySalamndr (approved):
LGTM

PR #348, opened 2025-03-02
combine env and runMode

ntseng:
Summary
-------
This commit combines the seqeulize `env` setting ("development" | "test" | "production") with our previous `runMode` setting ("development" | "migration" | "prod" | undefined). It also moves the model initializing logic into `bot.js` and labels the features of that file that are intended to have their own instance per shard.

This change was originally also planned to include making `models/index.js` into a more traditional index file, from which we could import model type data:
```
const { Bounty, Completion, User } = require ("./models");
```
however, something to do with Sequelize's base `Model` class is preventing VS Code from knowing the shape of `Model` children (defeating the purpose of exporting the models as types). With that and init logic being moved to `bot.js`, the `models/index.js` file is now empty (and has been removed).

Local Tests Performed
---------------------
- [x] /scoreboard command still fetches data from database
- [x] /item command correctly changes into and out of debug mode


source/bot.js
SillySalamndr, line 1: In this transition, we lost the `await` in a few key locations that was added to ensure that we had the database connection ready. Is there a particular reason why this was removed? Do we have some other guarantee of this?

ntseng: For the `await`s in the model file search, those are being handled by `return...then()` (L50 ~ L72). The ones in the event listeners are handled by the fact that those definitions are now in the last `.then()` of the chain. This is also a good reminder that I can remove some now unused `async`s from the event listeners.

SillySalamndr: Ah, that's really difficult to read in the diff, and also pretty counterintuitive in my opinion. There's a general practice to have your event code at the root layer if at all possible, since you want the system to _set up_ the event listeners as soon as is reasonable, even if the rest of the system isn't quite ready yet.

ntseng: Are we in a "not possible" situation? The whole reason we're being this picky about order is to ensure the database is set up before event listeners might try to use the database.

Still trying to accommodate "event listeners on root", dAPI login is the last thing done before the event listeners are defined. We could assume no events will be emitted for us before login to license moving the listeners to root (but this reasoning feels obfuscated to me).

SillySalamndr: There's absolutely no way we're in a "not possible" situation, since we were literally doing this on the last commit. It seems like you're not believing me, so I'm going to make a branch of this branch that demonstrates the change that I'd be preferring here.

SillySalamndr: Branch created with commit ID 28676cbe07bd8bd8b0ef6e2a1b21c3a5a4e8cedf.

ntseng: Noted, will sleep on it.

PR #356, opened 2025-03-03
create endpoint `findUserItemEntry()`

ntseng:
## NO REVIEW CANDIDATE

Summary
-------
- create endpoint `findUserItemEntry()`

Local Tests Performed
---------------------


SillySalamndr (approved):
LGTM

I agree with the "no review" tag on this one, but a bit cautiously. This feels like catch-up with getting certain baseline functionality made available in the logic layer, but my general instinct (without looking at the actual code review contents) would be that new logic layer functions and functionality would have code reviews, since (on average) that _should_ represent a new feature or something novel.

^ Putting these thoughts down once so I don't end up trying to repeat them across...dozens of reviews.

PR #357, opened 2025-03-03
rename class `Item` to `ItemTemplate` to disambiguate with table model `Item`

ntseng:
Summary
-------
- rename class `Item` to `ItemTemplate` to disambiguate with table model `Item`

Local Tests Performed
---------------------
- [x] Profile Colorizer still works

### DIFF
--- a/source/classes/ItemTemplate.js
+++ b/source/classes/ItemTemplate.js
@@ -1,7 +1,7 @@
 const { CommandInteraction } = require("discord.js");
 const { Sequelize } = require("sequelize");
 
-class Item {
+class ItemTemplate {
 	/**
 	 * @param {string} nameInput
 	 * @param {string} descriptionInput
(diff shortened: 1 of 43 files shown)


SillySalamndr (approved):
LGTM

This is maybe the kind of work that doesn't fit well with the idea of code reviews? The work here is functionally a find-and-replace, and as much as having a second set of eyes to confirm that there weren't any false positives might be a good idea some of the time, this is also the kind of changes where it is easy to miss false negatives. I have mixed feelings on that.

This is also a place where having the colorizer code condensed would make the review flow better, but I digress.

ntseng:
Good to get the difference in perspective out of the way here: though there's low risk of incorrect code or bad patterns here, I didn't mark this as "no review" to use reviews as a way to publicize the class name change.

PR #359, opened 2025-03-03
add endpoint `findCurrentServerGoal()`

ntseng:
## NO REVIEW CANDIDATE

Summary
-------
- add endpoint `findCurrentServerGoal()`

Local Tests Performed
---------------------

### DIFF
--- a/source/logic/goals.js
+++ b/source/logic/goals.js
@@ -12,6 +12,13 @@ function setDB(database) {
 	db = database;
 }
 
+/** *Finds the most recent ongoing Goal for the specified Company*
+ * @param {string} companyId
+ */
+function findCurrentServerGoal(companyId) {
+	return db.models.Goal.findOne({ where: { companyId }, state: "ongoing", order: [["createdAt", "DESC"]] });
+}
+
 /** *Queries for a Company's most recent Goal and the GP contributed to it*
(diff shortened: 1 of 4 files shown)


source/logic/goals.js
```diff
@@ -41,7 +48,7 @@ async function progressGoal(companyId, progressType, userId) {
 		goalCompleted: false,
 		contributorIds: []
 	};
-	const goal = await db.models.Goal.findOne({ where: { companyId, state: "ongoing" } });
+	const goal = await db.models.Goal.findOne({ where: { companyId, state: "ongoing" }, order: [["createdAt", "DESC"]] });
```
SillySalamndr, line 51: Was this added during this pull request, or is this a weird artifact of the order of pull requests or something? This change doesn't seem to belong here.

SillySalamndr (commented):
Overall LGTM, would like to double-check a change.

source/logic/goals.js
ntseng: This is an intended bugfix to that query sometimes fetching an out of date Goal. It's a duplicate of the `findCurrentServerGoal()` query, but I didn't want to solve the "reusable code in the logic files" problem in this pass.

SillySalamndr (approved):

PR #361, opened 2025-03-03
add endpoint `changeSeasonXP()`

ntseng:
Summary
-------
- add endpoint `changeSeasonXP()`
- fix skip DMing bot check always returning true

Local Tests Performed
---------------------
- [x] used `/moderation xp-penalty` to verify `model.increment()` works with negative values
- [x] bot no longer attempts to DM other bots when using `/moderation xp-penalty`


SillySalamndr (approved):
LGTM

PR #363, opened 2025-03-03
create endpoint `deleteCompanyParticipations()`

ntseng:
## NO REVIEW CANDIDATE

Summary
-------
- create endpoint `deleteCompanyParticipations()`

Local Tests Performed
---------------------

### DIFF
--- a/source/bot.js
+++ b/source/bot.js
@@ -224,7 +224,7 @@ client.on(Events.GuildDelete, async guild => {
 	db.models.Bounty.destroy({ where: { companyId: guild.id } });
 	db.models.Completion.destroy({ where: { companyId: guild.id } });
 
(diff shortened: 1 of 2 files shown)


SillySalamndr (approved):
LGTM - agree with your overall assessment of "no review" in spite of my comment below

The change is fine and good, but there's a decent chance that the functionality being created here is far too atomic to really matter. We should consider having a single logic layer function that does all of the guild delete functionality at once, even if that function repeats some database calls that would otherwise exist elsewhere.

ntseng:
Agreed on endpoints being too atomic for now, but am generally operating under the strategy of "make one-to-one or very close functionality mappings now to get `database` out of interfaces, then bundle them in a sensible way when they're all enumerated in the same file later".

PR #368, opened 2025-03-05
combine `env` and `runMode`

ntseng:
Summary
-------
- combine `env` and `runMode`
- create `dbReady` promise
- use || to default runMode to "development" in case of empty string
- fix global scope iterator in setDB iteration
- rename `sequelize` to `dbConnection`
- rename `client` to `dAPIClient`

Local Tests Performed
---------------------
- [x] `/scoreboard` still fetches from database
- [x] `/item` uses debug mode when `runMode` is not "production"
- [x] `/item` does not use debug mode when `runMode` is "production"


SillySalamndr (approved):
I think there's a minor concern that we only really care about "production" in the new code, but the "development" and "test" states are likely useful enough representations that we'll want to keep them for future use. Maybe we can capture run state in an enum to document this better in the code? That's not something that should be or needs to be addressed here.

ntseng:
Personal opinion: I like TS enums a lot more than JS enums, enough to the point that I'm willing to gamble on not making the enum until we need it in case we impelement TS before then.

SillySalamndr:
I mean, I think that's mostly due to JS enums basically not being enums in any reasonable way. By enforcing types, TS is able to represent the _most important part of enums in most languages_ which is basically impossible in JS. Even with how weak JS enums _are_ I think that they're at least valuable in giving a singular named location where all enum values are listed, instead of needing to those enum values listed in every relevant location. However, I don't think the value gain is enough to warrant doing it in all cases.

I'm gonna hit that squash and merge button.

PR #371, opened 2025-03-05
create endpoint `findRecentSecondings()`

ntseng:
## NO REVIEW CANDIDATE

Summary
-------
- create endpoint `findRecentSecondings()`

Local Tests Performed
---------------------


source/logic/toasts.js
```diff
@@ -11,6 +11,13 @@ function setDB(database) {
 	db = database;
 }
 
+/** *Find the Secondings of specified seconder for the purposes of Crit Toast and Rewarded Toast tracking*
+ * @param {string} seconderId
+ */
+function findRecentSecondings(seconderId) {
+	return db.models.Seconding.findAll({ where: { seconderId, createdAt: { [Op.gt]: new Date(new Date() - 2 * timeConversion(1, "d", "ms")) } } });
+}
```
SillySalamndr, line 19: hard coding the recency metric here feels odd, and i could see some logic around having different calling locations wanting a different idea of "recent". (off the top of my head, having different day ranges for crits and for stats pulled by moderators would be relatively reasonable.) being able to feed in "standard" configuration to this function might also be valuable in the future. this implementation is likely still a reasonable default, though.

idea: move the raw date value created into a function parameter, with the default value for that parameter as the current date value used in this function. we'd get better flexibility for this function in the future, while not increasing the complexity in the sole place we use this function.

```suggestion
/** *Find the Secondings of specified seconder for the purposes of Crit Toast and Rewarded Toast tracking*
 * @param {string} seconderId
 * @param {Date} createdDate
 * returns {Promise<db.models.Seconding[]>} All Secondings for the given ID that happen at or after the given date
function findRecentSecondings(seconderId, createdDate = new Date(new Date() - 2 * timeConversion(1, "d", "ms"))) {
	return db.models.Seconding.findAll({ where: { seconderId, createdAt: { [Op.gt]: createdDate } } });
}
```

SillySalamndr: The formatting on the returns type might be a bit off, but stating it in the JSDoc also gives a better place to describe the return output.

SillySalamndr (commented):
A potential suggestion that I think may be an improvement for future use, but doesn't really change the state of the code as it stands.

source/logic/toasts.js
SillySalamndr: Is there a response to this? Is this not desired?

ntseng: I made some comments, but didn't realize they were stuck in the shadow realm of "pending review mode". TLDR: Seems reasonable, but we wouldn't be using it yet and I'm trying not to expand or change functionality in the current refactor.

SillySalamndr: Fair enough. Do we want to make this an issue to track this? Otherwise, feel free to send the squash & merge.

ntseng: At this point, I'm pretty convinced we're going to want to do a polish pass after the instance handling refactor. I think that'll be easier with all the functionality (per interface) in the same files, and we might even have different opinions on the best way to bundle functionality when looking at it like that. That would put my personal opinion of "does this need an issue?" at "this most likely results as unnecessary backlog grooming", but wouldn't be opposed if you think there's a big risk we forget this specific issue.

SillySalamndr: In the spirit of "backlog as todo list so we don't forget" and "backlog grooming is unavoidable," I've made the issue #377 and linked the issue to this comment chain.

PR #372, opened 2025-03-05
create endpoint `findStaleToasteeIds()`

ntseng:
Summary
-------
- create endpoint `findStaleToasteeIds()`

Local Tests Performed
---------------------

### DIFF
--- a/source/logic/toasts.js
+++ b/source/logic/toasts.js
@@ -14,6 +14,8 @@ function setDB(database) {
 }
 
 /** *Get the ids of the rewarded Recipients on the sender's last 5 Toasts*
+ *
+ * Duplicated stale toastee ids are intended as a way of recording accumulating staleness
  * @param {string} senderId
  * @param {string} companyId
  */
--- a/source/logic/toasts.js
+++ b/source/logic/toasts.js
@@ -11,6 +11,17 @@ function setDB(database) {
 	db = database;
 }
 
+/** *Get the ids of the rewarded Recipients on the sender's last 5 Toasts*
+ * @param {string} senderId
+ * @param {string} companyId
+ */
+async function findStaleToasteeIds(senderId, companyId) {
+	const lastFiveToasts = await db.models.Toast.findAll({ where: { senderId, companyId }, include: db.models.Toast.Recipients, order: [["createdAt", "DESC"]], limit: 5 });
+	return lastFiveToasts.reduce((list, toast) => {
+		return list.concat(toast.Recipients.filter(reciept => reciept.isRewarded).map(reciept => reciept.recipientId));
+	}, []);
+}
+
 /**
  * @param {Guild} guild
  * @param {Company} company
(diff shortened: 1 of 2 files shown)


source/logic/toasts.js
```diff
@@ -11,6 +11,17 @@ function setDB(database) {
...
 
+/** *Get the ids of the rewarded Recipients on the sender's last 5 Toasts*
+ * @param {string} senderId
+ * @param {string} companyId
+ */
+async function findStaleToasteeIds(senderId, companyId) {
+	const lastFiveToasts = await db.models.Toast.findAll({ where: { senderId, companyId }, include: db.models.Toast.Recipients, order: [["createdAt", "DESC"]], limit: 5 });
+	return lastFiveToasts.reduce((list, toast) => {
+		return list.concat(toast.Recipients.filter(reciept => reciept.isRewarded).map(reciept => reciept.recipientId));
+	}, []);
```
SillySalamndr, line 22: The IDs must be unique, right? This is probably a place to use a set, or to force the entries to be unique.

A weird way of doing that, which preserves the Array typed output:
```suggestion
	return Array.from(new Set(lastFiveToasts.reduce((list, toast) => {
		return list.concat(toast.Recipients.filter(reciept => reciept.isRewarded).map(reciept => reciept.recipientId));
	}, [])));
```

A way of doing that which integrates the Set into the `reduce`:
```suggestion
	return Array.from(lastFiveToasts.reduce((set, toast) => {
		toast.Recipients.filter(reciept => reciept.isRewarded).map(reciept => reciept.recipientId).foreach(id => set.add(id));
		return set;
	}, new Set()));
```

SillySalamndr (changes requested):
There's a (minor) risk of duplicate IDs. I've added some suggestions that could address that.

source/logic/toasts.js
ntseng: Duplicated ids is intended, that way repeatedly toasting to the same hunter has accumulating staleness. I considered changing the return to `Record<userId, repeats>`, but felt the follow-up changes were out of scope.

SillySalamndr: I suppose this is a case where the new logic layer obfuscates intention a bit. I *do* think that something like you're suggesting would add some clarity (tracking repeats is intentional) without losing potential flexibility with this function (fetching stale toastees may be useful outside the original context), but I think we can get 90% of the way there with a more verbose JSDoc (either in the initial comment or with an @returns to describe the output directly).

SillySalamndr (approved):
A JSDoc update would help with the one point of confusion, but otherwise looks fine with expanded information.

source/logic/toasts.js
ntseng: The repeats map is more performant (less array searching) and concise as well, but even though I like it I'm waiting until the polish pass in case we find other functionality that would bundle well with it. Will update documentation for now.

PR #374, opened 2025-03-05
add findAll endpoints for seasons interface

ntseng:
Summary
-------
- `findSeasonParticipations()`
- `bulkFindParticipations()`
- `findHunterParticipations()`

Local Tests Performed
---------------------

### DIFF
--- a/source/commands/stats.js
+++ b/source/commands/stats.js
@@ -40,7 +40,7 @@ module.exports = new CommandWrapper(mainId, "Get the BountyBot stats for yoursel
 					const { xpCoefficient } = await logicLayer.companies.findCompanyByPK(interaction.guild.id);
 					const currentLevelThreshold = Hunter.xpThreshold(hunter.level, xpCoefficient);
 					const nextLevelThreshold = Hunter.xpThreshold(hunter.level + 1, xpCoefficient);
-					const participations = await database.models.Participation.findAll({ where: { userId: hunter.userId, companyId: hunter.companyId }, order: [["createdAt", "DESC"]] });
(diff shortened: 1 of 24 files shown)


source/commands/stats.js
```diff
@@ -40,7 +40,7 @@ module.exports = new CommandWrapper(mainId, "Get the BountyBot stats for yoursel
 					const { xpCoefficient } = await logicLayer.companies.findCompanyByPK(interaction.guild.id);
 					const currentLevelThreshold = Hunter.xpThreshold(hunter.level, xpCoefficient);
 					const nextLevelThreshold = Hunter.xpThreshold(hunter.level + 1, xpCoefficient);
-					const participations = await database.models.Participation.findAll({ where: { userId: hunter.userId, companyId: hunter.companyId }, order: [["createdAt", "DESC"]] });
+					const participations = await logicLayer.seasons.findHunterParticipations(hunter.userId, hunter.companyId);
```
SillySalamndr, line 43: By looking at where this function is used, it feels like it could be more "elegant" to just pass the "entire" hunter object here, and use destructured assignment in the parameters of the function to pull off only the `userId` and the `companyId` in the `seasons` logic file implementation of this function. That's neither here nor there though, and the intentionality here is its own kind of nice.

SillySalamndr (approved):
A minor comment, but overall looks good here.

source/commands/stats.js
ntseng: I noticed that possibility as well and didn't want to opt into a model fetch requiring a different model at this stage. If the `Participation` always needs the `Hunter` to be fetched first, we may want to do that with associations instead.

SillySalamndr: I can agree with that, yeah.

Issue #388, opened 2025-03-10
Reorganize `logicLayer.bounties` fetch functionality


ntseng:
In addition to the standard consolidation and extensibility improvements, the bounties interface also has the following dev headache to solve:
- The two branches of `findBounty()` represent not only different queries for fetching the same bounty, but different types of fetches in the first place. The `findByPk()` path is agnostic to bounty state, where as the `findOne()` path only fetches bounties in the "open" state. This may lead to unexpected results since the the different parameters now imply a whole use-case instead of a query type (eg future devs fetching by `userId`, `companyId`, and `slotNumber`, may not remember the detail of that path specifying "open" bounties only).

PR #389, opened 2025-03-10
implement `findBounty()` in more places

ntseng:
Summary
-------
- implement `findBounty()` in more places
- rename properties in object bountyInfo to `userId` and `companyId` to match entity name

Future Work
--------
- Discovered `findBounty()` is being called for two different use-cases, which unnecessarily increases implementation complexity, wrote up #388 to record

Local Tests Performed
---------------------


SillySalamndr (approved):
LGTM

PR #397, opened 2025-03-10
implement `getParticipantCount()` in Goal Initializer

ntseng:
Summary
-------
- implement `getParticipantCount()` in Goal Initializer

The existing code was originally incorrectly read, making it look like ties in placements would reduce the Participant count returned. In reality, ties are placed at the lowest placement they tie at (instead of "1, 2, 2", it would be "I, 3, 3"), which would give a correct count, but is still a more confusion way to achieve the count than the direct Participant count endpoint.

Local Tests Performed
---------------------

### DIFF
--- a/source/items/goal-initializer.js
+++ b/source/items/goal-initializer.js
@@ -17,7 +17,7 @@ module.exports = new ItemTemplate(itemName, "Begin a Server Goal if there isn't 
 		const eligibleTypes = ["bounties", "toasts", "secondings"];
 		const goalType = eligibleTypes[Math.floor(Math.random() * eligibleTypes.length)];
 		const previousSeason = await logicLayer.seasons.findOneSeason(interaction.guildId, "previous");
-		const activeHunters = previousSeason ? (await database.models.Participation.findOne({ where: { seasonId: season.id }, order: [["placement", "DESC"]] })).placement : 3;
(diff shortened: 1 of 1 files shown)


source/items/goal-initializer.js
```diff
@@ -17,7 +17,7 @@ module.exports = new ItemTemplate(itemName, "Begin a Server Goal if there isn't
 		const eligibleTypes = ["bounties", "toasts", "secondings"];
 		const goalType = eligibleTypes[Math.floor(Math.random() * eligibleTypes.length)];
 		const previousSeason = await logicLayer.seasons.findOneSeason(interaction.guildId, "previous");
-		const activeHunters = previousSeason ? (await database.models.Participation.findOne({ where: { seasonId: season.id }, order: [["placement", "DESC"]] })).placement : 3;
+		const activeHunters = previousSeason ? await logicLayer.seasons.getParticipantCount(previousSeason.id) : 3;
 		const requiredGP = activeHunters * 20;
```
SillySalamndr, line 21: Do we want to set a minimum `requiredGP`?

SillySalamndr (approved):
Changes overall look good, maybe moving the minimum requirement to the `requiredGP` is a good change in the future.

PR #403, opened 2025-03-13
create endpoint `findParticipationWithTopParticipationStat()`

ntseng:
Summary
-------
- create endpoint `findParticipationWithTopParticipationStat()`

Local Tests Performed
---------------------


SillySalamndr (approved):
Overall yes, I like this change. I think there's a small risk that's created here in that if any of these properties are renamed on the model, there's a good chance that it doesn't get updated here. How do we mitigate this? I...have no idea.

ntseng:
When we get TypeScript, I may make an enum out of the Participation properties, which can then be set as the parameter type. Then with the right scanning config, there could be a warning/error raised when the string provided doesn't match the enum anymore.

PR #406, opened 2025-03-14
remove database from interaction wrappers

ntseng:
Prerequisite Commit
------------
1. #400

Summary
-------
- Remove database from interaction wrappers

Local Tests Performed
---------------------
- [x] bot still turns on (no build errors, circular dependencies, or start-up errors)

PR #407, opened 2025-03-14
create class `SubcommandWrapper`

ntseng:
Prerequisite PRs
-------------
1. #400 
2. #406 

Summary
-------
- create class SubcommandWrapper

The 4k of changed lines may look intimidating, but the vast majority of that is indentation changes. Previously, every subcommand effect function was typed individually, now as a class there is better documentation on types and exact shape of data payload is abstracted out of the individual files.

There is a minor loss in type coverage as the `args` array is more rigid in its typing now, but I think that'll be fixable in TypeScript through template types.

Local Tests Performed
---------------------

### DIFF
--- a/source/bot.js
+++ b/source/bot.js
@@ -166,7 +166,7 @@ dAPIClient.on(Events.InteractionCreate, async interaction => {
 			return;
 		}
 
-		contextMenu.execute(interaction, dbConnection, runMode);
+		contextMenu.execute(interaction, runMode);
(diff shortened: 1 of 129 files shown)


SillySalamndr:
Git is doing the frustrating thing where the diffs aren't actually lining up, either, so it is very difficult to see that the only changes in some sections are spacing.

SillySalamndr (approved):
LGTM

Issue #409, opened 2025-03-16
Missing Season value on complete


SillySalamndr:
When attempting to complete a bounty, it will crash on not having a `season`. At some point, the season fetch was removed and should be added again before the `for` loop.

PR #410, opened 2025-03-16
#279 & #409: add time based item drop rate limiting

### LINKED ISSUE #279: Rate Limit Item Drops
Definitely do a flat quota: bonus drops can be premium

SillySalamndr:
Moved the roll code into a separate function to better manage item rolls in the logic layer, since this code (other than roll rate) should function the same regardless of the person being rolled for.

I've added some convenience functions for calculating dates which are not required to be done in this pull request, but were done here due to them being convenient.

Includes a fix for not pulling the `season` data.

Fixes #279

Summary
-------

Local Tests Performed
---------------------
- [X] item limit (silently) hit over repeated testing
- [X] item drop text still displays as relevant

Issue
-----
Closes #279

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,6 +13,8 @@ Server Goals are BountyBot objectives everyone on the server contributes to. A g
 - Fixed crash when using `/create-default rank-roles` twice in a row
 - Fixed `/moderation user-report` listing XP awarded for recent bounties as "undefined"
 - Fixed Dark Purple Profile Colorizer not dropping
+- Item drops are now limited to 2 items per day, or 4 items if you are a premium user
+
 ## BountyBot Version 2.8.0:
 ### Context Menu Options
 - Added the following BountyBot functionality to the Apps dropdown when right-clicking on a User or Message:
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -13,7 +13,7 @@ Server Goals are BountyBot objectives everyone on the server contributes to. A g
 - Fixed crash when using `/create-default rank-roles` twice in a row
 - Fixed `/moderation user-report` listing XP awarded for recent bounties as "undefined"
 - Fixed Dark Purple Profile Colorizer not dropping
-- Item drops are now limited to 2 items per day, or 4 items if you are a premium user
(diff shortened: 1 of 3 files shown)


source/logic/bounties.js
```diff
@@ -1,8 +1,9 @@
 const { Sequelize, Op } = require("sequelize");
-const { Guild, GuildMember } = require("discord.js");
+const { Guild, GuildMember, time } = require("discord.js");
```
ntseng, line 2:
```suggestion
const { Guild, GuildMember } = require("discord.js");
```
`time` looks like an accidental import here, but it's a cool formatter we might want to keep in mind in the future.
image

ChangeLog.md
```diff
@@ -13,6 +13,8 @@ Server Goals are BountyBot objectives everyone on the server contributes to. A g
 - Fixed crash when using `/create-default rank-roles` twice in a row
 - Fixed `/moderation user-report` listing XP awarded for recent bounties as "undefined"
 - Fixed Dark Purple Profile Colorizer not dropping
+- Item drops are now limited to 2 items per day, or 4 items if you are a premium user
```
ntseng, line 16:
```suggestion
- Item drops are now limited to 2 items per day, or 4 items for Premium users
```
The 2nd person tense sticks out a bit among the 3rd person passive technical documentation is usually written in (also, we should be confident that the feature offerings are indeed "Premium").

source/logic/bounties.js
SillySalamndr: This almost certainly happened with an unfortunate autocomplete with the new time calculation code.

ChangeLog.md
SillySalamndr: I don't know how you're reading a lack of confidence in the old wording, but I like the new wording better anyways.

ntseng:
Updated PR title and added #409 reference. Created #411 to track known bug.

ntseng (changes requested):
Looks good! Blocking on the accidental import not because it's awful, but because it's trivial to fix.

SillySalamndr:
Commits for the suggested changes should already be up.

ntseng (approved):

PR #412, opened 2025-03-17
#332: avoid crash if Bounty's embed's Completers field overflows

### LINKED ISSUE #332: Crash if Evergreen Bounty's embed's Completers field overflows
Because evergreen bounties accumulate completers, long running evergreen bounties risk overflowing the Completers field of their embed (can be viewed with `/evergreen list bounty-hunter: ${testbot mention}`). Embed field value character limit is recorded as `MAX_EMBED_FIELD_VALUE_LENGTH` in `constants.js`. While #329 is already planning to add some basic de-duplication to the field generation, the embed builder should also do the following:
- [ ] skip invalid bounty hunters (eg someone who completed the evergreen bounty but has since left the server)
- [ ] have logic for cutting off the field if too long without cutting mid user mention

While not in the scope of this issue, if ideas for the following arise during implementation, please record them:
- In the case that some completers are being cut off, what interface can we provide for users to look up the full completers list?
- Instead of only skipping invalid bounty hunters, can we clean up the database when finding invalidated Completions? Will we need to given "on leave delete records" behavior?

ntseng:
Summary
-------
-  avoid crash if Bounty's embed's Completers field overflows

Pivots
-------
Decided to pivot away from user validation: on bounty render is a fairly difficult place to predict for database updates to be happening and checking if a user is in a guild requires fetching the entire guild member (a dAPI "are these members in this guild?" endpoint would be ideal for that).

Also decided to do a full replacement of text to avoid the string modification problem that only really adds "completer order politics".

Local Tests Performed
---------------------
- [x] `/evergreen list` with too many completers doesn't crash due to completers field overflowing

Issue
-----
Closes #332

### DIFF
--- a/source/models/bounties/Bounty.js
+++ b/source/models/bounties/Bounty.js
@@ -1,7 +1,7 @@
 ﻿const { EmbedBuilder, Guild, ActionRowBuilder, ButtonBuilder, ButtonStyle, heading, userMention } = require('discord.js');
 const { Model, Sequelize, DataTypes } = require('sequelize');
 const { Company } = require('../companies/Company');
-const { SAFE_DELIMITER, MAX_MESSAGE_CONTENT_LENGTH } = require('../../constants');
+const { SAFE_DELIMITER, MAX_MESSAGE_CONTENT_LENGTH, MAX_EMBED_FIELD_VALUE_LENGTH } = require('../../constants');
 const { timeConversion, commandMention, listifyEN } = require('../../util/textUtil');
 const { Completion } = require('./Completion');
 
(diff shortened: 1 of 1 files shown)


SillySalamndr (approved):
Overall LGTM, and matches our other patterns. There's maybe space here to consider doing a short list (X, Y, and more!), but that would be paired with a pass on other fields to do similar.

SillySalamndr (approved):
Overall LGTM, and matches our other patterns. There's maybe space here to consider doing a short list (X, Y, and more!), but that would be paired with a pass on other fields to do similar.

SillySalamndr:
Oh, it did it twice, sorry.

PR #413, opened 2025-03-17
#308: migrate `showcaseBounty()` from utils to frontend

### LINKED ISSUE #308: Migrate `showcaseBounty()` from utils to logic and frontend layers

ntseng:
Summary
-------
- migrate `showcaseBounty()` from utils to frontend

Local Tests Performed
---------------------
- [x] no regressions in slash command showcase
- [x] no regressions in showcase item
- [x] no regressions in showcase button

Issue
-----
Closes #308


SillySalamndr:
Are we absolutely sure we want to do this? We're now repeating functionally the same code 3 times, in a case where we'd likely not want the behavior to be different across those implementations.

ntseng:
I think the pattern can be improved a lot as well, but how exactly isn't super obvious to me. As I was digging into the code, enough differences started popping out that I'm not sure where to consolidate any more.
| Functionality | Description | Layer | Variations |
| - | - | - | -|
| dAPI Validations | "Is BountyBot in the selected channel?" & "Is the user allowed to post in the selected channel?" | Frontend | How bounty is fetched |
| Bounty Validation | "Is this bounty still open?" | Logic | |
| Bounty and Poster Updates | increment "showcaseCount" and set poster's "lastShowcaseTimestamp" | Logic | Do not set "lastShowcaseTimestamp" on item version, so it's not subject to CD |
| Response/Acknowledgement | Send message and update bounty posting  | Frontend | How to fetch response channel |

The V shape this functionality makes in layers is part of what makes me reluctant to keep it coupled, in addition to the fact that both frontend steps are slightly different implementations (different inputs and outputs). Maybe the logic steps should have their own endpoints?

SillySalamndr:
Logic steps having independent endpoints might be a place to start, especially since we can call the logic layer an arbitrary number of times without notable processing penalty.

For this specific case, the more that I'm thinking about it, the important parts that we'd want to consolidate here are likely the shape of the rendering output, but not necessarily the sending of the response itself. If we were to consolidate exactly the process of creating (but not sending) the showcase, that could probably belong on the Bounty model under the pattern of "is a means to represent myself" that we've been using previously.

ntseng:
In that case, we're already done, I think (except for the endpoints)? The rendered object is just the bounty embed again.

SillySalamndr:
I- wow that's a lot more shared non-render-generating code than i was expecting then. I'm still instinctively trying to figure out a sane way of reducing the duplication here, but it might not be reasonable to (and was maybe unreasonable to condense this in the first place).

SillySalamndr:
It does seem odd that the bounty state is being checked here instead of in the logic layer (and results in doing a database call from the "frontend" code through the `reload` call), but moving exactly and only that to the logic layer seems excessive in this case, too.

It was easy to miss due to the sea of green, but yeah, there's also a (really reasonable) behavior difference there as well, and this new flow allows for the difference to exist without needing conditional flow.

I guess this is more proof that de-duplicating the code flow _has_ benefits, but seeing them immediately is pretty tough, and the de-duplication is still shocking. (Even with the "inherent" drive to de-duplicate removed from many years of practice, it still shows up as a reaction at times.)

SillySalamndr (approved):
LGTM

SillySalamndr:
An idea for future, but not now: consolidating specifically validation to a separate class/set of `validator` files/classes/functions might be the kind of de-duplication that would work, but that kind of concept warrants a more generalized and methodical pass.

ntseng:
> An idea for future, but not now: consolidating specifically validation to a separate class/set of `validator` files/classes/functions might be the kind of de-duplication that would work, but that kind of concept warrants a more generalized and methodical pass.

I like this idea a lot. Not totally clear on exactly what it'd look like, but that's what brainstorming's for.

PR #414, opened 2025-03-17
#345: consolidate Colorizer items

### LINKED ISSUE #345: Merge Colorizer code from Colorizer branch in a way that makes sense*
A branch has been made that attempts to merge the self-similar code from across all colorizers so that the code should (hopefully) be easier to manage. This code is not perfectly up to date and is merely an idea for what could be done to reduce redundancy across the colorizers (which only truly differ by color).

Goals:

1. Update the colorizer branch with the current state of the "average" colorizer.
2. Attempt to improve what is in that branch.
3. Merge the new solution or reject it.

ntseng:
Summary
-------
- consolidate Colorizer items

Decided to make a new branch because the old branch was stale (was still using `database`, etc) and only at a prototype level of polish (included syntax errors, outdated documentation, and unused imports).

Local Tests Performed
---------------------
- [x] using colorizers doesn't crash the bot

Issue
-----
Closes #345


SillySalamndr (commented):
I have mixed feelings on this. This obviously implements a kind of change that I'm a fan of, but does so by also implementing a new object type that only really functions as a wrapper for an array. Why not just use an array here instead?

SillySalamndr:
I'm not fully convinced that that is the correct representation of the concept being provided, though, and at least not enough of one for what is otherwise a small optimization. Are item templates intended to be independent objects on instantiation, or are they always conceptually bound to others that are similar, even after they are created? This distinction likely doesn't matter here for the foreseeable future, but it is the kind of choice that influences future code, and a model that doesn't tightly fit the idea being created is usually a place where there is code churn at best and bugs at worst. It is _sufficient_ to implement the desired code, though. I'm going to approve.

SillySalamndr (approved):

PR #432, opened 2025-03-24
#426: fix false positive on existing Server Goal check in Goal Initializer

### LINKED ISSUE #426: Goal Initializer false positive on onging Server Goal
When checking if there's currently an ongoing Server Goal, the Goal Initializer item will early out even if the most recent Goal has been completed.

ntseng:
Summary
-------
- fix false positive on existing Server Goal check in Goal Initializer
- fix completed goals being shown on scoreboards
- remove clamping to requiredGP on GP gain updates
   - clamping was previously added because having the existing GP be above the required GP "looked buggy", but a realistic case where the GP could be above the threshold was found during testing: if type doubling or another feature increased the GP gain above the remaining quota. in this case, it recognizes hunter contributions better if it shows the over-contribution

Local Tests Performed
---------------------
- [x] Goal Initializer works when a server has a completed goal
- [x] scoreboards don't show completed goals
- [x] over-contribution shown on toast embed

Issue
-----
Closes #426

### DIFF
--- a/source/buttons/bbcomplete.js
+++ b/source/buttons/bbcomplete.js
@@ -82,7 +82,7 @@ module.exports = new ButtonWrapper(mainId, 3000,
 						if (goalUpdate.gpContributed > 0) {
 							const { goalId, requiredGP, currentGP } = await logicLayer.goals.findLatestGoalProgress(interaction.guildId);
 							if (goalId !== null) {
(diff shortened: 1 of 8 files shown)


SillySalamndr (approved):
LGTM

PR #435, opened 2025-03-24
#431: fix premium check for item drop limit

### LINKED ISSUE #431: Premium item drop limit should check through constants instead of on db
`User.isPremium` is placeholder and not actually set in the current schema.

ntseng:
Summary
-------
- fix premium check for item drop limit

Local Tests Performed
---------------------

Issue
-----
Closes #431


SillySalamndr (approved):
LGTM

I'm contemplating if we actually want some means to update the value on the User table at minimum on startup, but we've also discussed that briefly as a "it'd be nice if we can skip this step."

Issue #427, opened 2025-03-23
`/raffle by-rank` doesn't find all eligible entrants


ntseng:
When running a Bronze Rank or up raffle, an entrant who was managable by the bot was skipped.

ntseng:
Changes in #429 are in conflict with changes for this fix (and should make this fix easier).

ntseng:
Rank a Bronze and Up raffle on the following season:
![Image](https://github.com/user-attachments/assets/51d7b8dc-7e82-47f1-a5c7-cf8a6e2292f8)

Got the following entrants ("Qualified" = not rank disqualified, "Eligible" = Qualified + manageable):
![Image](https://github.com/user-attachments/assets/0e0748a9-da2a-4a38-941d-eeeecfcdc6f4)

Permissions are as following:
![Image](https://github.com/user-attachments/assets/2cf8a08d-7e0c-43d5-91fe-ab9e70ab42c4)

Maybe both #428 and this bug report were human error in tracking a user's qualification status?

ntseng:
Moving to dismissed on "cannot reproduce".

PR #438, opened 2025-03-24
#429: add early out if rank goes missing during `/raffle by-rank`

### LINKED ISSUE #429: Crash in `/raffle by-rank` if rank is deleted between command start and resolution
The `findAllRanks` isn't sufficient to prevent that crash if it fails to find the specified rank.

ntseng:
Summary
-------
- catch Sequelize error on non-existent rank reload during `/raffle by-rank`
- move "nonemoji rank emoji" error handling to a DM, as the ephemeral message that would have held the error message gets cleaned up

This is a blocker for #427, which will be working in the same lines of code.

Had to change from `rankIndex` as select value to `varianceThreshold` to be able to detect missing ranks. This is likely pointing at changing a column on Hunter (in v2.10.0), given how volatile `rankIndex` is.

Local Tests Performed
---------------------
- [x] `/raffle by-rank` early outs instead of crashing if selected rank was deleted
- [x] `/raffle by-rank` can proceed beyond rank existence check

~~I'll need help testing the racing on this one. Also, I looked, but couldn't find any documentation on what Sequelize does if we try to reload an instance who's been deleted.~~ Sequelize throws an error (unhelpfully only containing the vague name "SequelizeInstanceError") when reloading a non-existant entity.

Issue
-----
Closes #429

### DIFF
--- a/source/commands/raffle/byrank.js
+++ b/source/commands/raffle/byrank.js
@@ -49,11 +49,15 @@ module.exports = new SubcommandWrapper("by-rank", "Select a user at or above a p
 				company.update("nextRaffleString", null);
 			});
 		}).catch(error => {
+			if (error.code === DiscordjsErrorCodes.InteractionCollectorError) {
+				return;
+			}
+
 			if (error.name === "SequelizeInstanceError") {
 				interaction.user.send({ content: "A raffle by ranks could not be started because there was an error with finding the rank you selected." });
 			} else if (Object.values(error.rawError.errors.data.components).some(row => Object.values(row.components).some(component => Object.values(component.options).some(option => option.emoji.name._errors.some(error => error.code == "BUTTON_COMPONENT_INVALID_EMOJI"))))) {
 				interaction.user.send({ content: "A raffle by ranks could not be started because this server has a rank with a non-emoji as a rankmoji." });
-			} else if (error.code !== DiscordjsErrorCodes.InteractionCollectorError) {
+			} else {
 				console.error(error);
 			}
 		}).finally(() => {
--- a/source/commands/raffle/byrank.js
+++ b/source/commands/raffle/byrank.js
@@ -20,7 +20,7 @@ module.exports = new SubcommandWrapper("by-rank", "Select a user at or above a p
 							const option = {
 								label: rank.roleId ? guildRoles.get(rank.roleId).name : `Rank ${index + 1}`,
 								description: `Variance Threshold: ${rank.varianceThreshold}`,
-								value: index.toString()
+								value: rank.varianceThreshold.toString()
 							};
 							if (rank.rankmoji) {
 								option.emoji = rank.rankmoji;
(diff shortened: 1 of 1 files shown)


ntseng:
Think I've figured out a way to test solo.

source/commands/raffle/byrank.js
```diff
@@ -29,31 +30,34 @@ module.exports = new SubcommandWrapper("by-rank", "Select a user at or above a p
...
 		}).catch(error => {
-			if (Object.values(error.rawError.errors.data.components).some(row => Object.values(row.components).some(component => Object.values(component.options).some(option => option.emoji.name._errors.some(error => error.code == "BUTTON_COMPONENT_INVALID_EMOJI"))))) {
-				interaction.reply({ content: "A raffle by ranks could not be started because this server has a rank with a non-emoji as a rankmoji.", flags: [MessageFlags.Ephemeral] });
-			} else if (error.code !== DiscordjsErrorCodes.InteractionCollectorError) {
+			if (error.code === DiscordjsErrorCodes.InteractionCollectorError) {
+				return;
+			}
+
+			if (error.name === "SequelizeInstanceError") {
+				interaction.user.send({ content: "A raffle by ranks could not be started because there was an error with finding the rank you selected." });
```
SillySalamndr, line 57: I think we can suggest the user takes an action here. Something like "Please try again." should go a long way.

source/commands/raffle/byrank.js
```diff
@@ -29,31 +30,34 @@ module.exports = new SubcommandWrapper("by-rank", "Select a user at or above a p
...
-				collectedInteraction.reply(`The winner of this raffle is: ${winner}`);
-				logicLayer.companies.findCompanyByPK(interaction.guild.id).then(company => {
-					company.update("nextRaffleString", null);
-				});
-			})
+		}).then(message => message.awaitMessageComponent({ time: 120000, componentType: ComponentType.StringSelect })).then(async collectedInteraction => {
+			const varianceThreshold = Number(collectedInteraction.values[0]);
+			const reloadedRanks = await Promise.all(ranks.map(rank => rank.reload()));
+			const rankIndex = reloadedRanks.findIndex(rank => rank.varianceThreshold === varianceThreshold);
+			const rank = ranks[rankIndex];
```
SillySalamndr, line 38: Since we're using the `rankIndex` from the `reloadedRanks`, shouldn't we be finding the rank by the reference in `reloadedRanks` instead? If the lists _are_ different there's a strong chance that the rank at that index is different in the new list.

SillySalamndr:
```suggestion
			const rank = reloadedRanks[rankIndex];
```
for easy fix application

SillySalamndr (changes requested):
I think that the new code overall accomplishes the goal, and changing this function to async does clean up the flow a bit with the new database/logic layer accesses. I think a potential bug has slipped through, but it is a very easy fix, and there's maybe a line of text that could be a bit better with the user-facing prompt.

ntseng:
@SillySalamndr Ready for re-review!

SillySalamndr (approved):
LGTM

PR #452, opened 2025-04-02
update sequelize to v6.37.7

ntseng:
Summary
-------
- update sequelize to v6.37.7

Local Tests Performed
---------------------
- [x] according to Sequelize's release notes, the two patches were a documentation update and an update to handling BLOB type (which we don't use any of)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
 		"": {
 			"dependencies": {
 				"discord.js": "^14.18.0",
-				"sequelize": "^6.37.5",
+				"sequelize": "^6.37.7",
 				"sequelize-cli": "^6.6.2",
 				"sqlite3": "^5.1.7"
 			}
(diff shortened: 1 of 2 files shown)


SillySalamndr:
Is this slated for the bug patch?

SillySalamndr:
Ah, wait, draft. Not slated for the bug patch.

This appears to be a minor version upgrade, which should be safe overall. Is there a particular motivator for this outside of being up-to-date with the major version that we're currently operating on? There also should be another major version above this one, if I'm remembering correctly. Are we wanting to push for that upgrade?

ntseng:
Minor patch motivated by being up to date. Sequelize 7, is still in alpha and is looking like a full overhaul on their interfaces, so I'm looking for a signal that their interface has solidified first.

SillySalamndr (approved):
LGTM

PR #453, opened 2025-04-04
SillySalamndr/issue448

### LINKED ISSUE #448: Refactor `/rank info` to show all ranks
If message would overflow, send a text file.

SillySalamndr:
Implemented `/rank info` listing all ranks instead of needing to specify a rank by variance. This implementation includes a means to send the ranks as a text file, even through hitting character limits to trigger this seems unlikely due to a (default) rank limit of 25.

Fixed a bug found along the way where creating the default ranks would fail due to referencing a renamed function that removes all old ranks.

Local Tests Performed
---------------------
- [X] `/rank info` displays all ranks available from the default set when the bot is configured with that initial set
- [X] `/rank info` displays all 25 ranks after creating 25 ranks to be displayed

Issue
-----
Closes #448

### DIFF
--- a/source/commands/create-default/rankroles.js
+++ b/source/commands/create-default/rankroles.js
@@ -16,7 +16,7 @@ module.exports = new SubcommandWrapper("rank-roles", "Create the default ranks f
 				console.error(error);
 			}
 		});
-		const deletedCount = await logicLayer.ranks.deleteRanks(interaction.guild.id);
+		const deletedCount = await logicLayer.ranks.deleteCompanyRanks(interaction.guild.id);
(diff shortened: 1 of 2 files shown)


ntseng (approved):
LGTM! Good find on the outdated logic method name.

For context on the "unlikely to overflow the message", looking at it now I generally agree that it's fairly unlikely to overflow in this case. However, there is still an edge case where it can happen, as snowflakes actually grow in length over time (so yeah, probably in the far future). The combination of "uncaught dAPI message overflow error is a crash", "this will make later implementations of #456 more consistent" (ie "if message has variable length, always apply overflow handler function"), and "it's already done" seem like good enough post-hoc justification to me.

PR #454, opened 2025-04-04
SillySalamndr/issue447

### LINKED ISSUE #447: Delete Discord Role when using `/remove rank`
Leftover roles (especially simiar looking ones) can be a pain to clean up.

SillySalamndr:
Added code in the 3 main locations where ranks are removed to also remove associated roles. Removing a singular rank removes the associated role, if that role existed. Bulk rank removal collects a list of the role IDs, removes empy/invalid entries, and then deletes each role by that ID reference. (There is no bulk role removal provided by Discord, unfortunately.) While it may make sense to try to move this code to the logic layer (since this action should alway happen when removing one or more ranks), this interaction with Discord likely doesn't belong on that layer, so we have a minor amount of duplication.

Local Tests Performed
---------------------
- [X] Removing single rank removes associated role (visible changes in old messages)
- [X] Removing all ranks through `create-default` removes associated roles
- [X] Removing the bot removes rank roles

Issue
-----
Closes #447

### DIFF
--- a/source/bot.js
+++ b/source/bot.js
@@ -252,6 +252,10 @@ dAPIClient.on(Events.GuildDelete, async guild => {
 	logicBlob.seasons.deleteCompanySeasons(guild.id);
 	logicBlob.goals.deleteCompanyGoals(guild.id);
 	logicBlob.hunters.deleteCompanyHunters(guild.id);
+	logicBlob.ranks.findCompanyRanks(guild.id)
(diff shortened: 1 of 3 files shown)


ntseng (approved):
LGTM!

I've got some refactor plans cooking for providing a place for "shared frontend non-model code", which we can flesh out more during our meeting triaging #391 .

PR #457, opened 2025-04-04
fix unknown interaction in bounty board complete

ntseng:
Summary
-------
This was found while investigating the April 2nd Unknown Interaction, but I don't think it's the same one. This one happens because `interaction` is acknowledged twice (once if the bounty isn't found, again on clean up), but the one being looked for is on `collectedInteraction` (the payload being sent is the rewards message).

I also wasn't able find the original Unknown Interaction.

Local Tests Performed
---------------------


SillySalamndr (approved):
LGTM

PR #458, opened 2025-04-07
decompose `Hunter.addXP()`

ntseng:
Summary
-------
- decompose `Hunter.addXP()` to split up frontend and logic layer responsibilities
- refactor `Hunter.maxSlots()` (now `static Hunter.getBountySlots()`) to calculate for any level (decouple Hunter instances)
- refactor toast seconding to batch critical toasts
- refactor `/bounty list` to have more formal handling of evergreen listing
- fix Epic and Legendary XP Boost items responding that they are normal XP Boost items
- reload `company` and `hunter` in `/bounty post` instead of re-fetching fresh
- refactor to calculate `Hunter.level` during runtime as `Hunter.getLevel()`
- refactor to calculate `Company.xp` during runtime as `Company.getXP()`
- refactor to calculate `Company.level` during runtime as `Company.getLevel()`
- fix `/toast` rewards not being affected by festivals (despite frontend reporting it)

Local Tests Performed
---------------------
- [x] migration removes level columns
- [x] migration reversion restores level columns
- [x] `/evergreen complete` under 100x festival still completes with both hunter level up and company level up notes
- [x] `/toast` still completes


migrations/20250407182747-virtualLevels.js
```diff
@@ -0,0 +1,13 @@
...
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+	async up(queryInterface, Sequelize) {
+		await queryInterface.removeColumn("Hunter", "level");
+		await queryInterface.removeColumn("Company", "xp");
+		await queryInterface.removeColumn("Company", "level");
+	},
+	async down(queryInterface, Sequelize) {
+		console.log("WARNING virtualLevels migration not reversible because it deletes columns");
```
SillySalamndr, line 11: This isn't true, right? We should be able to query for the virtual fields and then calculate the resulting xp/level to re-insert. Now, this is a _lot_ of work, but if we're dedicated on the reversibility of migrations we can implement it here.

ntseng: oh, true. i'll probably save that for tomorrow.

source/models/companies/Company.js
```diff
@@ -33,9 +34,40 @@ class Company extends Model {
...
+		return {
+			open: this.openBountyThumbnailURL ?? "https://cdn.discordapp.com/attachments/545684759276421120/734093574031016006/bountyboard.png",
+			complete: this.completedBountyThumbnailURL ?? "https://cdn.discordapp.com/attachments/545684759276421120/734092918369026108/completion.png",
+			deleted: "https://cdn.discordapp.com/attachments/545684759276421120/734093574031016006/bountyboard.png"
+		};
+	}
+
 	festivalMultiplierString() {
 		if (this.festivalMultiplier != 1) {
-			return ` ***x${this.festivalMultiplier}***`;
```
SillySalamndr, line 38: The `x` is no longer a part of the bolded and italicized text. Is this intentional?

ntseng: Nope, good catch

migrations/20250407182747-virtualLevels.js
SillySalamndr, line 1: I know that I'm the one that suggested the `down` function to actually exist, but I think we can do better with this functionality in a couple of ways:
1. For overall clarity of the function, I'd suggest doing all structure changes next to each other. (So, in this case, both checks for adding the `level` column at the top of the function.) The way that I'm thinking about it: the structure changes are prerequisites of the data changes, and as much as those two structure changes are independent of each other, getting the dependencies done first frees up mental capacity to understand the data changes that follow.
1. If we query for the company table information at about the same time as the hunter information, we can build the company xp coefficient map in its own loop which will greatly simplify the `for` loop for the hunter data update. Additionally, the xp coefficient map construction would no longer need to have a check for if the company is already in the map.
1. Abstracting away the level calculation as a separate function would help with the readability of the functionality within the SQL string. I understand the calculations are _slightly_ different, but it should be reasonable to divide the parameter input by 3 for the same effect.
1.

migrations/20250407182747-virtualLevels.js
```diff
@@ -0,0 +1,38 @@
...
+		const companyTableDescription = await queryInterface.describeTable("Company");
+		if (!("level" in companyTableDescription)) {
+			await queryInterface.addColumn("Company", "level", { type: Sequelize.INTEGER, defaultValue: 1 });
+		}
+		const [companies] = await queryInterface.sequelize.query("SELECT * FROM Company;");
+		for (const company of companies) {
+			const [result] = await queryInterface.sequelize.query(`SELECT SUM(level) FROM Hunter WHERE companyId IS ${company.id};`);
+			await queryInterface.sequelize.query(`UPDATE Company SET level = ${Math.floor(Math.sqrt(result[0]["SUM(level)"] / 3) + 1)} WHERE id IS ${company.id}`);
+		}
+	}
```
SillySalamndr, line 37:
```suggestion
		//region structure
		const hunterTableDescription = await queryInterface.describeTable("Hunter");
		if (!("level" in hunterTableDescription)) {
			await queryInterface.addColumn("Hunter", "level", { type: Sequelize.BIGINT, defaultValue: 1 });
		}
		const companyTableDescription = await queryInterface.describeTable("Company");
		if (!("level" in companyTableDescription)) {
			await queryInterface.addColumn("Company", "level", { type: Sequelize.INTEGER, defaultValue: 1 });
		}
		//endregion
		//region data
		const [hunters] = await queryInterface.sequelize.query("SELECT * FROM Hunter;");
		const [companies] = await queryInterface.sequelize.query("SELECT * FROM Company;");
		const companyXPCoefficientMap = {};
		for (const company of companies) { // Set up XP map
			companyXPCoefficientMap[company.id] = company.xpCoefficient;
		}
		for (const hunter of hunters) { // Set hunter levels
			await queryInterface.sequelize.query(`UPDATE Hunter SET level = ${Math.floor(Math.sqrt(hunter.xp / companyXPCoefficientMap[hunter.companyId]) + 1)} WHERE userId is ${hunter.userId} AND companyId is ${hunter.companyId}`);
		}

		for (const company of companies) { // Set company levels
			const [result] = await queryInterface.sequelize.query(`SELECT SUM(level) FROM Hunter WHERE companyId IS ${company.id};`);
			await queryInterface.sequelize.query(`UPDATE Company SET level = ${Math.floor(Math.sqrt(result[0]["SUM(level)"] / 3) + 1)} WHERE id IS ${company.id}`);
		}
		//endregion
	}
```

migrations/20250407182747-virtualLevels.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
```
SillySalamndr, line 2:
```suggestion
function xpToLevel(xp, xpCoefficient = 3) {
	return Math.floor(Math.sqrt(xp / xpCoefficient) + 1);
}
```

migrations/20250407182747-virtualLevels.js
SillySalamndr: I've added suggestions to implement the things above, but I would like to note that what you're doing _is_ good and on read looks like it should work.

SillySalamndr (commented):
Things look good overall. I have some suggestions for the migration, two thirds of which fall into style rather than functionality. The middle suggestion of that set should be more performant, but that is a relatively minor concern at the scale we're working at. Take what you want out of those suggestions, and let me know which changes if any make your cut.

SillySalamndr (approved):
LGTM

PR #465, opened 2025-04-11
#417: switch `/toast` option from String to User

### LINKED ISSUE #417: Replace "parse User Mentions from command option string" with user select command option
In `/bounty complete` (and other places) we use a string option on the command to allow users to mention an arbitrary number of users in the input. However, mentions in string options don't autocomplete or parse on mobile, so we should switch to the specific user option.

ntseng:
Summary
-------
- switch `/toast` option from String to User
- added run mode to bot startup logging
- added validation for rejecting bountybot banned users from receiving toasts
- added `getCompanyHunterMap()` endpoint to get normalized map to all of a company's hunters

Future Work
-------------
Split `/bounty add-completers` #466, `/bounty complete` #467, `/bounty remove-completers` #468, and `/evergreen complete` #469 to separate tickets due to unexpected scope for `/toast`.

Local Tests Performed
---------------------
- [x] `/toast` still resolves without regression
- [x] `/toast` rejects BountyBot banned toastees

Issue
-----
Closes #417


source/bot.js
```diff
@@ -253,9 +253,9 @@ dAPIClient.on(Events.GuildDelete, async guild => {
 	logicBlob.goals.deleteCompanyGoals(guild.id);
 	logicBlob.hunters.deleteCompanyHunters(guild.id);
 	(await logicBlob.ranks.findAllRanks(guild.id))
-			.map(r => r.roleId)
-			.filter(id => !!id)
-			.forEach(id => guild.roles.delete(id, 'Cleaning up BountyBot roles during kick.'));
+		.map(r => r.roleId)
+		.filter(id => !!id)
+		.forEach(id => guild.roles.delete(id, 'Cleaning up BountyBot roles during kick.'));
```
SillySalamndr, line 258: Any particular reason for this style change? I tend to do double indentation for chained operations on a new line to make them distinct from indentation for blocks (like `if` blocks and `for` blocks).

source/commands/toast.js
```diff
@@ -112,16 +120,40 @@ module.exports = new CommandWrapper(mainId, "Raise a toast to other bounty hunte
...
+		name: "fourth-toastee",
+		description: "The fourth bounty hunter to whom you are raising a toast",
+		required: false
+	},
+	{
+		type: "User",
+		name: "fifth-toastee",
+		description: "The fifth bounty hunter to whom you are raising a toast",
+		required: false
+	},
```
SillySalamndr, line 156: Oh, we're needing to hard-code these slots...I'm not sure I'm a huge fan of needing to do this, but an alternative that would allow for any number of toastees (recursive modal requests until the user returns nothing) would also be annoying. Not sure there's anything we can do about that.

SillySalamndr: Either that solution, or Discord giving use real multi-selects...which is somehow unlikely.

source/commands/toast.js
```diff
@@ -22,22 +23,29 @@ module.exports = new CommandWrapper(mainId, "Raise a toast to other bounty hunte
...
+		const bannedIds = [];
+		const validatedToasteeIds = [];
+		for (const optionalToastee of ["toastee", "second-toastee", "third-toastee", "fourth-toastee", "fifth-toastee"]) {
+			const guildMember = interaction.options.getMember(optionalToastee);
+			if (guildMember) {
+				if (hunterMap[guildMember.id]?.isBanned) {
+					bannedIds.push(guildMember.id);
+					continue;
+				}
+				if (runMode !== "production" || (!guildMember.user.bot && guildMember.user.id !== interaction.user.id)) {
```
SillySalamndr, line 35: I know that I've personally used `continue` in a number of places and it can be useful, but I'd probably prefer using `else if` here since the logic here doesn't continue without this second check..

SillySalamndr (commented):
Overall logic and process looks sound, but I have a question and an optional request.

source/bot.js
ntseng: Auto-format on save tbh

SillySalamndr (approved):

PR #470, opened 2025-04-11
#461: make deleted bounty thumbnail customizable

### LINKED ISSUE #461: Add ability to customize deleted bounty embed thumbnails
Should be a migration, a new column on the Company table, and an update to `Company.getThumbnailURLMap()`

ntseng:
Summary
-------
- make deleted bounty thumbnail customizable
- remove old migrations

This functionlaity shouldn't be accessible for normal usage (when bounties are deleted, their bounty board thread is deleted instead of updated). Extended the functionality in case we end up making a bounty look up function in the future.

Local Tests Performed
---------------------
- [x] migration up adds column to database
- [x] migration down removes column from database
- [x] customized bounty thumbnail is used on bounty board
   - tested this by using the new column for the open state instead of the deleted state (since bounty embeds never show with a deleted state)

Issue
-----
Closes #461


SillySalamndr (approved):
LGTM

PR #473, opened 2025-04-14
#467: switch `/bounty complete` hunter options from String to User

### LINKED ISSUE #467: Switch `/bounty complete` option from String to User
The String option doesn't work on mobile. This will involve refactoring the command to remove `extractUserIdsFromMentions()`. Note that `extractUserIdsFromMentions()` filters out the interacting user's id, which will need to be readded after removing that function call.

ntseng:
Summary
-------
- switch `/bounty complete` option from String to User
- refactor initial member collection to use Collection instead of two Arrays
   - Collection uses Map's key-based `.has()`, so it should be as performant as an Array and a Set

Local Tests Performed
---------------------
- [x] `/bounty complete` resolves when not given hunter options
- [x] `/bounty complete` credits hunters given by hunter options

Issue
-----
Closes #467


SillySalamndr:
Looks good so far

SillySalamndr (approved):
LGTM

PR #475, opened 2025-04-14
#469: switch `/evergreen complete` hunter options from String to User

### LINKED ISSUE #469: Switch `/evergreen complete` option from String to User
The String option doesn't work on mobile. This will involve refactoring the command to remove `extractUserIdsFromMentions()`. Note that `extractUserIdsFromMentions()` filters out the interacting user's id, which will need to be readded after removing that function call.

ntseng:
Summary
-------
- switch `/evergreen complete` `hunters` option from String to User
- remove deduplication from hunter list to allow awarding for multiple turn-ins in same command

Local Tests Performed
---------------------
- [x] `/evergreen complete` still resolves with only required hunter
- [x] `/evergreen complete` still resolves with an optional hunter

Issue
-----
Closes #469

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -3,6 +3,7 @@
 ### User Options in Slash Commands
 Slash commands originally accepted users as a string (instead of Discord's user filtering) to allow users to mention as many users as they wanted. However, this doesn't work on mobile, so those slash commands have been remade to use Discord's user filtering.
 - `/toast` now accepts 1 required toastee and up to 4 optional ones. The `message` option has been changed to the first option to group the toastee options.
+- `/evergreen complete` now requires 1 hunter and up to 4 optional ones. Duplicate hunters are now accepted in case the hunter has multiple turn-ins to be awarded for.
 ### Other Changes
 - Added `/seasonal-ranks`, which allows all bounty hunters to look up the server's list of seasonal ranks (removed `/rank info` which was only usable by Premium users)
 - Fixed BountyBot banned users being able to receive toasts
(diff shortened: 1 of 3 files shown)


SillySalamndr:
Looks good so far

ChangeLog.md
```diff
@@ -5,6 +5,7 @@ Slash commands originally accepted users as a string (instead of Discord's user
 - `/toast` now requires 1 toastee and up to 4 optional ones. The `message` option has been changed to the first option to group the toastee options.
 - `/bounty add-completers` has been renamed to `/bounty verify-turn-in` and now requires 1 bounty hunter and up to 4 optional ones.
 - `/bounty complete` now accepts up to 5 optional hunters.
+- `/evergreen complete` now requires 1 hunter and up to 4 optional ones. Duplicate hunters are now accepted in case the hunter has multiple turn-ins to be awarded for.
```
SillySalamndr, line 8: These might be getting repetitive. We might want to do a pass before release to group these together so we can simply list the name changes, followed by which commands received this change.

source/util/textUtil.js
```diff
@@ -156,23 +156,6 @@ function dateInFuture(timeMap) {
...
-	const ids = [];
-	let results;
-	while ((results = idRegExp.exec(mentionsText)) != null) {
-		const id = results[1];
-		if (!excludedIds.includes(id)) {
-			ids.push(id);
-		}
-	}
-	return ids;
-}
```
SillySalamndr, line 174: On one hand, I'd not be surprised if a function like this becomes useful at a later point and we feel the need to revive this. On the other, I kinda hope that it doesn't become the case.

SillySalamndr (approved):
LGTM. There is one possible point of cleanup to do closer to release, but that shouldn't be done here.

PR #476, opened 2025-04-14
avoid duplicating toastees

ntseng:
Summary
-------
- `/toast` filters out duplicate toastees
- `/toast` provides a followup on successful toasts if any toastees were banned

Local Tests Performed
---------------------
- [x] `/toast` filters out duplicate toastees
- [x] `/toast` provides a followup on successful toasts if any toastees were banned

### DIFF
--- a/source/commands/toast.js
+++ b/source/commands/toast.js
@@ -39,7 +39,7 @@ module.exports = new CommandWrapper(mainId, "Raise a toast to other bounty hunte
 		let bannedText;
 		if (bannedIds.size > 1) {
 			bannedText = `${listifyEN([...bannedIds.values()].map(id => userMention(id)))} were skipped because they're banned from using BountyBot on this server.`;
-		} else if (bannedIds.length === 1) {
+		} else if (bannedIds.size === 1) {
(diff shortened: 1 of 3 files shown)


source/commands/toast.js
```diff
@@ -30,7 +30,7 @@ module.exports = new CommandWrapper(mainId, "Raise a toast to other bounty hunte
 			if (guildMember) {
 				if (hunterMap[guildMember.id]?.isBanned) {
 					bannedIds.push(guildMember.id);
-				} else if (runMode !== "production" || (!guildMember.user.bot && guildMember.user.id !== interaction.user.id)) {
+				} else if (runMode !== "production" || (!guildMember.user.bot && guildMember.user.id !== interaction.user.id && !validatedToasteeIds.includes(guildMember.id))) {
 					validatedToasteeIds.push(guildMember.id);
```
SillySalamndr, line 34: This is likely a good check to do, but we may also be interested in having `validatedToasteeIds` and/or `bannedIds` be `Set`s instead to avoid this kind of check. Also worth noting that this kind of issue can apply to the `bannedIds` list, too, but it is likely causes less issues there than on the validated IDs.

SillySalamndr (approved):
LGTM

PR #478, opened 2025-04-16
#346: remove `logicLayer` from scoreboard embed generator parameters

### LINKED ISSUE #346: Move Company/Season find-or-create calls into "command" code for embed creation
Currently we have logic layer calls within the util functions that generate seasonal and overall bounty boards. This isn't necessarily bad, but it is causing some cases where we're doing find-or-create calls for the same seasons or companies multiple times in the program flow, when it would feel like an opportune place to simply pass a single company/season through the code flow.

Found during review on #310

ntseng:
Summary
-------
- remove `logicLayer` from scoreboard embed generator parameters

Local Tests Performed
---------------------
- [x] `/scoreboard seasonal` resolves
- [x] `/scoreboard overall` resolves

Issue
-----
Closes #346


source/models/companies/Company.js
```diff
@@ -170,15 +165,15 @@ class Company extends Model {
...
-	async overallScoreboardEmbed(guild, logicLayer) {
-		const hunters = await logicLayer.hunters.findCompanyHuntersByDescendingXP(guild.id);
+	async overallScoreboardEmbed(guild, hunters, ranks) {
 		const hunterMembers = await guild.members.fetch({ user: hunters.map(hunter => hunter.userId) });
-		const rankmojiArray = (await logicLayer.ranks.findAllRanks(guild.id)).map(rank => rank.rankmoji);
+		const rankmojiArray = ranks.map(rank => rank.rankmoji);
 
 		const scorelines = [];
-		for (const hunter of hunters) {
+		for (const hunter of hunters.sort((a, b) => b.xp - a.xp)) {
```
SillySalamndr, line 176: Something worth considering here: now that this is sorting by XP, we can `break` out of the loop once we hit the first hunter with 0 XP. This kind of micro-optimization would belong in a different task though, and may hurt readability through code bloat.

SillySalamndr (approved):
LGTM

source/models/companies/Company.js
ntseng: Refactoring the immediately following `xp > 0` check is equivalent complexity, so I've gone ahead and done that.

PR #490, opened 2025-05-02
#446: `/rank remove` improvements

### LINKED ISSUE #446: `/rank remove` improvements
- Add confirmation button
- Add more details about the rank to delete

ntseng:
Summary
-------
- `/rank remove` accepts input from a select
- `/rank remove` can remove ranks in bulk
- `/rank remove` has a confirmation button
- `/rank remove` recalculates ranks

Local Tests Performed
---------------------
- [x] `/rank remove` can remove ranks in bulk
- [x] `/rank remove` recalculates ranks

Issue
-----
Closes #446


ntseng:
Shared categories of `dAPIRequests` and `messageParts` don't appear to be useful categorizations. In finding a place for the new `disabledSelectRow()` function, it naturally felt like that would belong in `messageParts`, but that required importing `truncateTextToLength()` from `dAPIRequests`, which caused a circular dependency.

Further, similar functions `bountesToSelectOptions()` and `rankArrayToSelectOptions()` live in `dAPIRequests` and `messageParts` respectively.

source/frontend/shared/dAPIRequests.js
```diff
@@ -36,7 +37,25 @@ function bountiesToSelectOptions(bounties) {
...
+		const option = {
+			label: rank.roleId ? allGuildRoles.get(rank.roleId).name : `Rank ${index + 1}`,
+			description: `Variance Threshold: ${rank.varianceThreshold}`,
+			value: rank.varianceThreshold.toString()
+		};
+		if (rank.rankmoji) {
+			option.emoji = rank.rankmoji;
+		}
+		return option;
+	}).slice(0, SelectMenuLimits.MaximumOptionsLength);
```
SillySalamndr, line 58: I'm a bit concerned with truncating this list in this way, at least without telling the user about it. This issue would be an extreme edge case, though, and probably should be put into its own issue.

SillySalamndr (approved):
LGTM, one possible future issue that can probably be addressed separately (and maybe in a future where it actually comes up)

source/frontend/shared/dAPIRequests.js
ntseng: This is here as last minute crash prevention, though it should never actually function. It's enforced on rank creation that servers aren't allowed to have more than 25 ranks (because it would crash selects).

PR #491, opened 2025-05-02
#456: create validation function `contentOrFileMessagePayload()`

### LINKED ISSUE #456: Consolidate "variable length message overflow handling"
There are several places where we send messages with lengths that are based on things like "number of bounty hunters" or other user submitted content. These places often employ logic to instead attach a text file if the message would be too long, since getting an unexpected dAPI error back for a message overflow would be a crash for us. When we create a place for shared frontend code, consolidating that logic should reduce that duplication.

ntseng:
Summary
-------
- consolidate "variable length message overflow handling"

Local Tests Performed
---------------------
- [x] `/seasonal-ranks` resolves as usual when below character limit
- [x] `/seasonal-ranks` sends a file when above character limit

Issue
-----
Closes #456


ntseng:
Emoji as plain text get interpreted as a handful of unrelated characters. Gonna say creating a mapping function that "finds a unicode equivalent" and "handles uploaded discord emoji" to run the text through beforehand is out of scope.

SillySalamndr (approved):
LGTM

PR #502, opened 2025-05-09
#333 & #460: refactor rank and placement calculation

### LINKED ISSUE #333: Move `getRankUpdates()` out of utils
The function's dependence on a database connection indicates it's a better fit for the logic layer (probably seasons).

### LINKED ISSUE #460: Add `Season.xpStandardDeviation` and refactor `Hunter.rank` to `Participation.xpVariance`
Goal: make finding a hunter's rank less difficult and the database schemas less obfuscated to understand

ntseng:
Summary
-------
- replace `frontend/shared/toBeMoved.getRankUpdates()` with `logicLayer.seasons.updateCompanyPlacementsAndRanks()`
- fixed typo "particiaptionProperty" to "participationProperty"

Paradigm Shifts
---------------
- Moving `rank` from Hunter to Participation allows `updateCompanyPlacementsAndRanks()` to function with only the current Season's Participations, rather than all of the Company's Hunters AND the current Season's Participations. This means the frontend store can shift to passing Participation maps around instead of Hunter maps.
- Using `Map`s instead of `Object`s for map data structures. This PR is leveraging `Map` functionality of being iterable and having a size property (instead of needing to `Object.keys().length`).
- `Participation.rankIndex` was originally going to be refactored to a rank's id, but needs to be kept as an index to encode rank order (so we can detect "rank up" vs "rank down"). Efforts will be made to formalize "descending" as the official ordering.

Local Tests Performed
---------------------
- [ ] bot still turns on (no build errors, circular dependencies, or start-up errors)

Test plan tbw

Issue
-----
Closes #333 
Closes #460

### DIFF
--- a/source/database/models/seasons/Participation.js
+++ b/source/database/models/seasons/Participation.js
@@ -9,6 +9,10 @@ class Participation extends Model {
 			foreignKey: "companyId"
 		})
 	}
+
+	nextRankXP(season, descendingRanks) { //TODONOW replace Hunter.nextRankXP usages
+		return Math.ceil(season.xpStandardDeviation * descendingRanks[this.rankIndex].varianceThreshold + season.xpMean - this.xp);
+	}
 }
 
 /** @param {Sequelize} sequelize */
@@ -46,6 +50,9 @@ function initModel(sequelize) {
 		placement: {
 			type: DataTypes.INTEGER,
 			defaultValue: 0
+		},
+		rankIndex: {
+			type: DataTypes.INTEGER
 		},
 		postingsCompleted: {
 			type: DataTypes.INTEGER,
--- a/source/logic/seasons.js
+++ b/source/logic/seasons.js
@@ -1,4 +1,5 @@
 const { Sequelize, Op } = require("sequelize");
+const { Participation, Rank, Season } = require("../database/models");
 
 /** @type {Sequelize} */
 let db;
@@ -48,6 +49,19 @@ function getDQCount(userId, companyId) {
 	return db.models.Participation.sum("dqCount", { where: { userId, companyId } }) ?? 0;
 }
 
+/** *Returns a Map of userId to Participation for all Participations in the specified Season*
+ * @param {string} companyId
+ */
+async function getCompanyParticipationMap(seasonId) {
+	/** @type {Record<string, Participation} */
+	const participationMap = new Map();
+	const participations = await db.models.Participation.findAll({ where: { seasonId } });
+	for (const participation of participations) {
+		participationMap.set(participation.userId, participation);
+	}
+	return participationMap;
+}
+
 /** *Get all of a Season's Participations*
  * @param {string} seasonId
  */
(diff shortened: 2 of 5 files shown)


source/database/models/seasons/Participation.js
```diff
@@ -9,6 +10,31 @@ class Participation extends Model {
...
+		if (participations.size < 1) {
+			return null;
+		}
+		const mean = Season.calculateXPMean(participations);
+		let squareSum = 0;
+		for (const [_, participation] of participations) {
+			squareSum += (participation.xp - mean) ** 2;
+		}
+		return Math.sqrt(squareSum / participations.size);
+	}
```
SillySalamndr, line 37: Is there a particular reason for these functions to live here instead of in a file in one of our `shared` directories? These functions seem to not have any particular association to a Participation instance, and it doesn't appear to be a means to generate a representation of one or more `Participation` objects and instead is a means to generate a very narrow summary across multiple instances.

source/frontend/commands/moderation/xp-penalty.js
```diff
@@ -14,8 +14,10 @@ module.exports = new SubcommandWrapper("xp-penalty", "Reduce a bounty hunter's X
 		hunter.decrement({ xp: penaltyValue });
 		hunter.increment({ penaltyCount: 1, penaltyPointTotal: penaltyValue });
 		const [season] = await logicLayer.seasons.findOrCreateCurrentSeason(interaction.guildId);
-		logicLayer.seasons.changeSeasonXP(member.id, interaction.guildId, season.id, penaltyValue * -1);
-		getRankUpdates(interaction.guild, logicLayer);
+		logicLayer.seasons.changeSeasonXP(member.id, interaction.guildId, season.id, penaltyValue * -1); //TODONOW check for async problems
```
SillySalamndr, line 17: TODONOW?

source/frontend/commands/moderation/take-down.js
```diff
@@ -42,8 +42,10 @@ module.exports = new SubcommandWrapper("take-down", "Take down another user's bo
 				logicLayer.hunters.findOneHunter(posterId, interaction.guild.id).then(async poster => {
 					poster.decrement("xp");
 					const [season] = await logicLayer.seasons.findOrCreateCurrentSeason(interaction.guildId);
-					logicLayer.seasons.changeSeasonXP(posterId, interaction.guildId, season.id, -1);
-					getRankUpdates(interaction.guild, logicLayer);
+					logicLayer.seasons.changeSeasonXP(posterId, interaction.guildId, season.id, -1); //TODONOW check for async problems
```
SillySalamndr, line 45: TODONOW?

source/frontend/shared/messageParts.js
```diff
@@ -368,6 +368,7 @@ async function overallScoreboardEmbed(company, guild, hunters, ranks, goalProgre
 		if (hunter.xp < 1) {
 			break;
 		}
+		//TODONOW need participation for rankIndex?
```
SillySalamndr, line 371: TODONOW?
I do somewhat think that assigning the rank index to a participation record is an odd choice. The rank is of each hunter, right? At least conceptually, the rank index belongs to the hunter (at least so that there is only one relevant rank per hunter). Now, as far as how that influences the new implementation...it might be messier than is worth handling.

source/frontend/commands/moderation/take-down.js
ntseng: TODONOW is my personal tag for stuff remaining to do in the current PR, so yeah, this PR isn't done yet. (Not "TODO" to avoid also picking up other TODOs left in the codebase long term, but doesn't get spotted until I have a PR that isn't done in one shot.)

source/database/models/seasons/Participation.js
ntseng: Good suggestion, I'll move them to either `logic/shared` if the frontend doesn't use them or `root/shared` both logic and frontend use them.

source/frontend/shared/messageParts.js
ntseng: The Participation is the season-based stats per hunter and ranks reset each season, so I think it actually makes more sense to have the rank on the Participation. Having the ranks on Participation allows us to remove a cleanup step in the season reset (the former L70 of `/season-reset`), and opens functionality where we can tell hunters what their rank was in previous seasons. While this does create the possible bug of "wrong rank because fetched the wrong Participation", the query to ensure we get the responsibility of "ensure the fetched Participation is for the current Season" is well handled by the logic layer (if not already done).

source/logic/seasons.js
```diff
@@ -82,28 +90,133 @@ function findFirstPlaceParticipation(companyId, seasonId) {
...
+			updatePayload.rankIndex = index;
+			if (id in results) {
+				results[id].newRankIndex = index;
+				results[id].rankIncreased = isIncrease;
+			} else {
+				results[id] = { newRankIndex: index, rankIncreased: isIncrease };
+			}
+		}
+		await participationMap.get(id).update(updatePayload);
+	}
```
SillySalamndr, line 155: So, I'm assuming that there are IDs that can be shared between both `placementChanges` and `rankChanges`, right? merging these lists to then process them at least partially independently feels weird, but it does appear that this process lets you construct the `updatePayload` in a way that is maybe easier. I'd maybe argue that if we can have `rankChanges` and `placementChanges` in different loops it may make the code less complex by not needing to do the additional check and making both processes more independent.

SillySalamndr (approved):
Overall looks, good, with one question on complexity.

source/logic/seasons.js
ntseng: That read is correct. I'm gunning to have a later refactor that takes `Record<userId, MapOfAllHunterRewards>` (expanding `MapOfAllHunterRewards` to include level-ups, level-up rewards, and drops) and converts it into the equivalent array of strings (or maybe concatenates them in that same function?). This would encapsulate all of this search logic into its own function that only has to be run once per flow, where I think we'd be justified to have the complexity (rather than having it distributed here and other places). In addition, separating responsibilities, this will make it easier for us to make changes to reward string formatting and arrangement (eg maybe rank ups should be sub-bullets under the xp gains).

PR #504, opened 2025-05-17
#349 & #477: add Cooldowns for User Interactions and Items

### LINKED ISSUE #349: Create centralized location for cooldown management
Command cooldowns are per shard at the moment, and item cooldowns are somewhat hard to find, being located in the `/item` command file.

The shard manager would be a good candidate for this, as we don't want cooldown management instanced per shard, except for the fact that it would create a circular dependency with `bot.js` then needing to import from the shard manager.

### LINKED ISSUE #477: Track item cooldowns in database
The database can serve as a single source of truth for item cooldowns so that shards aren't racing to update cooldown timestamps.

See also: #349

SillySalamndr:
- Fixed the DataTypes reference on `interactionName` for both the migration and the model
- Fixed the command dictionary style references messed up in the transition
- Fixed some shared resource references messed up in the transition

Summary
-------

Local Tests Performed
---------------------
- [ ] bot still turns on (no build errors, circular dependencies, or start-up errors)
- (how did you ensure these changes don't break the code base and work as intended?)

Issue
-----
Closes #477
Closes #349


source/logic/cooldowns.js
```diff
@@ -0,0 +1,57 @@
...
+			isOnGeneralCooldown: false,
+			isOnCommandCooldown: true,
+			cooldownTimestamp: thisInteractions.cooldownTime,
+			lastCommandName: interactionName
+		};
+	}
+}
+
+async function updateCooldowns(userId, interactionName, interactionTime, interactionCooldown) {
+	const [interaction, wasCreated] = await db.models.UserInteraction.findOrCreate({ where: { userId, interactionName } });
```
ntseng, line 34:
```suggestion
	const [interaction, wasCreated] = await db.models.UserInteraction.findOrCreate({ where: { userId, interactionName }, defaults: { hitTimes: 1 } });
```
Adding this defaults option will make it so this `UserInteraction` is created with a `hitTimes` value of 1. Alternatively, I think we could also make the `defaultValue` of `hitTimes` 1 in the schema, since I can't think of a time we'd want to initialize a row with a `hitTimes` value of 0.

source/database/models/users/UserInteraction.js
```diff
@@ -0,0 +1,45 @@
...
+		interactionTime: {
+			type: DataTypes.DATE
+		},
+		lastInteractTime: {
+			type: DataTypes.DATE
+		},
+		cooldownTime: {
+			type: DataTypes.DATE
+		},
+		hitTimes: {
```
ntseng, line 34: This seems like it'll be very useful for us learning which commands are used most!

Seeing it like this has made me realize this won't be a good schema for doing an impact based cooldown scaling system (eg "each interaction increases global cooldown by 30 seconds until gcd stacks fall off"), which we might be interested in to clamp down on spam cases more precisely. Something like that may require a schema that has one row per interaction and we query for a time window to count rows. Probably best to discuss the value of that feature on voice before making changes here though.

source/frontend/commands/_commandDictionary.js
```diff
@@ -52,3 +52,17 @@ exports.setLogic = function (logicBlob) {
...
 	}
 }
+
+exports.updateCooldownMap = function(map) {
+	for (const commandKey in commandDictionary) {
+		map[commandKey] = commandDictionary[commandKey].cooldown;
+	}
+}
+
+exports.updatePremiumList = function(list) {
```
ntseng, line 62: This is good encapsulation, we might be able to refactor away from `getCommand()` exposing entire command objects soon.

This will also be a good place for us to have SKU and other dAPI premium logic in the future: https://discord.js.org/docs/packages/discord.js/14.19.3/CommandInteraction:Class#sendPremiumRequired

migrations/20250502193643-createUserInteraction.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+const { DataTypes } = require('sequelize');
+
+/** @type {import('sequelize-cli').Migration} */
+module.exports = {
+	async up(queryInterface, Sequelize) {
```
ntseng, line 6: With respect to your concerns about migration conflicts earlier: since it's only touching it's own new table, Sequelize should be ok to apply this before or after any of the other migrations we have in the works. No name timestamp meddling required!

ntseng (commented):
Direction is good! Main comments are some Sequelize tips and a extensibility question on `UserInteraction`'s schema.

source/logic/cooldowns.js
ntseng: Alternatively alternatively, unconditionally incrementing `hitTimes` trades a database ping for a logic simplification.

SillySalamndr: The idea was to default to 0 so that it is easier to do future math regarding added cooldown time with additional hits (multiplying additional cooldown time by number of hits on that command).

migrations/20250502193643-createUserInteraction.js
SillySalamndr: I still really hate this conceptually, since the "receipt" for the migration (the timestamp baked into the file name) lies about the expected application order of migrations.

source/database/models/users/UserInteraction.js
SillySalamndr: I'm...not quite following on the last paragraph. The whole point of having the hit time tracking is for the impact-based cooldown scaling (at least for individual users). This system isn't intended for long-term tracking yet, if that's what you're trying to get at, and I'm not sure we want something like that with this database schema (too easy to have this table bloat with rows for that).

source/index.js
```diff
@@ -146,13 +159,48 @@ dAPIClient.on(Events.ClientReady, () => {
...
+		return;
+	}
+	if (isOnCommandCooldown) {
+		interaction.reply({ content: `Please wait, \`/${interaction.commandName}\` is on cooldown. It can be used again <t:${cooldownTimestamp}:R>.`, flags: [MessageFlags.Ephemeral] });
+		return;
+	}
+	logicBlob.cooldowns.updateCooldowns(interaction.user.id, interaction.commandName, commandTime, cooldownMap[interaction.commandName]);
+	if (itemCommands.contains(interaction.commandName)) {
+		const itemName = interaction.options.getString("item-name");
+		const {isOnCommandCooldown, cooldownTimestamp, lastCommandName} = logicBlob.cooldowns.checkCooldownState(interaction.user.id, itemName, commandTime);
```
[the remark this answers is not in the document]
SillySalamndr: This implementation was initially assuming potentially separate cooldown states for the literal `item` command and the items themselves. Do we not want that?

package.json
```diff
@@ -8,6 +8,7 @@
 	"dependencies": {
 		"@sapphire/discord.js-utilities": "^7.3.2",
 		"discord.js": "^14.18.0",
+		"node-cron": "^3.0.3",
```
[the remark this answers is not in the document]
SillySalamndr: Yes, for the automatic database culling of the stale `UserInteraction` table entries (with a cooldown that has expired over a day ago, as of last discussion).

source/logic/cooldowns.js
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Check cooldown information based on the given user, intraction, and interaction time.
+ * @param {string} userId 
+ * @param {string} interactionName 
+ * @param {Date} interactionTime 
+ * @returns {{isOnGeneralCooldown: boolean, isOnCommandCooldown: boolean, cooldownTimestamp?: Date, lastCommandName?: string}}
+ */
+async function checkCooldownState(userId, interactionName, interactionTime) {
+	const allInteractions = await db.models.UserInteraction.findOne({ where: { userId }, order: [[ "cooldownTime", "DESC" ]] });
+	const gcdCooldown = !allInteractions ? new Date(0) : new Date(allInteractions.lastInteractTime.getTime() + GLOBAL_COMMAND_COOLDOWN); // TODO this sometimes returns an invalid date. Why?
```
SillySalamndr, line 21: The TODO might be due to bad data created by test states...but just in case, is there anything that seems wrong here?

ntseng (commented):
Did some bug consulting, will need more repro info to confirm.

source/logic/cooldowns.js
SillySalamndr: The "Invalid Date" was also in the database, which isn't a good sign. I'd bet that'd evaluate to `undefined` with `getTime()` pretty easily. I'm going to shelve this until the more pressing issue of the `interactionName` is figured out.

source/logic/cooldowns.js
```diff
@@ -0,0 +1,88 @@
...
+			isOnCommandCooldown: false,
+			cooldownTimestamp: gcdCooldown,
+			lastCommandName: allInteractions.interactionName
+		};
+	}
+	return checkCommandCooldownState(userId, interactionName, interactionTime);
+}
+
+async function checkCommandCooldownState(userId, interactionName, interactionTime) {
+	console.log(interactionName); // TODO find why this is `undefined` for cooldown checks coming from `commands/item.js`
```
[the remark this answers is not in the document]
SillySalamndr: It was typed `undefined`, and still is after adding the missing `await`.

ntseng (changes requested):
Pulled down to my local environment to see the cooldowns working for myself and ran into a handful of bugs.

## Migration leaves database in state that crashes on `findOrCreateBountyHunter()`
image
Steps to repro:
1. Migrate from existing database (pre-migration database file provided as `db_base.txt` as the `.sqlite` extension isn't supported by github, don't forget to change the extension back) [db_base.txt](https://github.com/user-attachments/files/20984514/db_base.txt)
2. Run `/about`

## Global Cooldown Check Doesn't Trigger
Reinitializing the database from scratch avoids the above crashes. I did that, then attempted to use `/about` several times in the same 2 seconds. They all succeeded.

## `UserInteraction.hitCount` not incremented
When checking the database after the previous test, I noticed the `hitCount` column was still at 0.

SillySalamndr:
> Pulled down to my local environment to see the cooldowns working for myself and ran into a handful of bugs (made sure to install `node-chron` and migrate).
> 
> ## Migration leaves database in state that crashes on `findOrCreateBountyHunter()`
> image
> Steps to repro:
> 1. Migrate from existing database (pre-migration database file provided as `db_base.txt` as the `.sqlite` extension isn't supported by github, don't forget to change the extension back) [db_base.txt](https://github.com/user-attachments/files/20984514/db_base.txt)
> 2. Run `/about`

I swear I fixed this already...and if this keeps coming up like this, I'm not sure it is solvable.

PR #516, opened 2025-05-30
#330: refactor `logicLayer.bounties.addCompleters()`

### LINKED ISSUE #330: Decouple mass fetch requests from single logic file function for addCompleter
Split from discussions from #299

ntseng:
Summary
-------
- rename to `checkTurnInEligiblity()`
- remove fetches unneeded for own logic
- move `bulkCreateCompletions` call to frontend
- use Sets instead of Arrays for built-in deduplication and faster search

Other Changes:
- switch `buildBountyEmbed()` to require Set instead of an Array (that is later converted to a Set)
- switch `updatePosting()` to require a Set for `buildBountyEmbed()`
- more "Bounty Credit" to "Turn-Ins" rewords

Local Tests Performed
---------------------
- [x] Context Menu Option (renamed to Record Bounty Turn-In) still records turn-ins
- [x] `/bounty record-turn-ins` still records turn-ins
- [x] Bounty Board Button: Record Turn-Ins still records turn-ins

Issue
-----
Closes #330


source/logic/bounties.js
```diff
@@ -80,69 +80,56 @@ function findBountyCompletions(bountyId) {
...
- * @param {string} userId
- * @param {string} companyId
+/** *Get a Set with the userIds of the specified Bounty's hunters*
+ * @param {string} bountyId
  */
-function findHuntersLastFiveBounties(userId, companyId) {
-	return db.models.Bounty.findAll({ where: { userId, companyId, state: "completed" }, order: [["completedAt", "DESC"]], limit: 5, include: db.models.Bounty.Completions });
+async function getHunterIdSet(bountyId) {
+	const completions = await db.models.Completion.findAll({ where: { bountyId } });
+	return completions.reduce((set, completion) => set.add(completion.userId), new Set());
```
SillySalamndr, line 88: Worth bringing up as a discussion: while I generally agree with your use of `reduce` here, this is also something that some devs would consider to be a "bad" use of reduce, since we change output types and do a process that is more "transformative" as a result. This can also be done pretty easily using a `map` and feeding the result into the constructor of a new `Set`.

ntseng: That just sounds like dogma convincing people to run an extra n iterations to me.

SillySalamndr: You're...not entirely wrong. I think there's some merit in more extreme situations to back off of reduce, but I also don't think this is anywhere near that extreme.

PR #517, opened 2025-05-30
#497: fix flows where evergreen bounty board wasn't updating

### LINKED ISSUE #497: `/evergreen complete` doesn't update bounty board with new hunters
The embed for the evergreen bounty isn't updated with the new bounty hunters after `/evergreen complete` (just not implemented). Make "updateEvergreenThread()" to abstract that functionality out and apply it across all `/evergreen` commands.

ntseng:
Summary
-------
- fix `/evergreen complete` not updating evergreen embed with new hunters
- fix `/evergreen swap` removing turn-in list from embed
- fix `/evergreen take-down` removing turn-in lists on all embeds
- fix `/festival start` not updating evergreen embeds with multiplier tags
- fix `/festival close` not removing multiplier tags from evergreen embeds
- rename `generateBountyBoardThread()` to `createEvergreenBountyThread()`

Design Changes?
---------------
The festival commands should technically also update the multiplier tags on all open user bounties as well as the evergreen bounties. However, that seems like it's a lot more than we bargained for, and only a minor improvement (users are probably already hyped about the xp multiplier). Maybe the correct call is to avoid putting the multiplier tag in the bounty board for open bounties in general?

Local Tests Performed
---------------------
- [x] `/evergreen complete` updates bounty board thread embed with new hunters
- [x] `/evergreen swap` no longer removes turn-in list from embed
- [x] `/evergreen take-down` no longer removes turn-in lists on all embeds
- [x] `/festival start` updates evergreen embeds with multiplier tags
- [x] `/festival close` removes multiplier tags from evergreen embeds

Issue
-----
Closes #497


SillySalamndr (commented):
One question on how we're handling server goals in one place, but otherwise looks good.

SillySalamndr (approved):

PR #526, opened 2025-06-27
#520: switch from Hunter arrays to maps in frontend

### LINKED ISSUE #520: Switch from Hunter[] to Map<string, Hunter> in frontend
Lots of frontend flows use a Hunter array to hold a company's hunters or other collections of hunters for data processing because Sequelize `findAll()`s return an array. Packing the entities into a Map first would allow hash based lookups (instead of iteration based searching) and still allow iteration with low risk of those hunter collections needing to be serialized to JSON.

Also investigate if reloading individual entities from a Map propogates the reload into the Map itself (obsoleting the need for a method for updating the Map).

ntseng:
Summary
-------

Local Tests Performed
---------------------

Issue
-----
Closes #520

### DIFF
--- a/source/database/models/companies/Company.js
+++ b/source/database/models/companies/Company.js
@@ -31,14 +31,18 @@ class Company extends Model {
 		})
 	}
 
-	/** @param {Hunter[]} hunters */
-	getXP(hunters) {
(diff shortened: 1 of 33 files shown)


SillySalamndr (changes requested):
I think there's a quick suggestion that'll make the resulting code a bit more clear.

SillySalamndr (approved):

PR #529, opened 2025-07-19
SillySalamndr/issue377

### LINKED ISSUE #377: Add recency metric to `findRecentSecondings`
Issue created from code review discussion https://github.com/Imaginary-Horizons-Productions/BountyBot/pull/371#discussion_r1982720231

When trying to move some code to the logic layer, an instance where a function could be made more flexible to more accurately cover current/future use cases was discovered. Moving the recency date (or equivalent) to a parameter in `findRecentSecondings` could allow this function to be reused for various server metrics instead of only being used by the seconding code itself.

There may be other locations like this, and a pass on marking those during this process would be appreciated, but not necessary. Any additional cases like this should have their own tasks.

SillySalamndr:
Fixes #377

Summary
-------
`findRecentSecondings` was made more generically applicable by allowing for the function to take in the amount of time in the past to limit by. Most all of the work in this task comes from additions to the JSDoc. There should be no functional changes from this task.

I apologize for not having a thorough test before making this pull request; I need to have a second user so that it is possible to have someone hit the seconding button.

Local Tests Performed
---------------------
- [ ] TODO: Group test to hit this function via toast + seconding

Issue
-----
Closes #377

### DIFF
--- a/source/logic/toasts.js
+++ b/source/logic/toasts.js
@@ -12,9 +12,16 @@ function setDB(database) {
 
 /** *Find the Secondings of specified seconder for the purposes of Crit Toast and Rewarded Toast tracking*
  * @param {string} seconderId
- */
-function findRecentSecondings(seconderId) {
-	return db.models.Seconding.findAll({ where: { seconderId, createdAt: { [Op.gt]: dateInPast({ d: 2 }) } } });
+ * @param {object} recency How far in the past to look for recent secondings. By default, 2 days.
+ * @param {number} recency.w An amount of time to look back in weeks.
+ * @param {number} recency.d An amount of time to look back in days.
+ * @param {number} recency.h An amount of time to look back in hours.
+ * @param {number} recency.m An amount of time to look back in minues.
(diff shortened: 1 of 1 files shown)


source/logic/toasts.js
```diff
@@ -12,9 +12,16 @@ function setDB(database) {
 
 /** *Find the Secondings of specified seconder for the purposes of Crit Toast and Rewarded Toast tracking*
  * @param {string} seconderId
- */
-function findRecentSecondings(seconderId) {
-	return db.models.Seconding.findAll({ where: { seconderId, createdAt: { [Op.gt]: dateInPast({ d: 2 }) } } });
+ * @param {object} recency How far in the past to look for recent secondings. Defaults to 2 days.
```
ntseng, line 15: I think using the [Partial TypeScript utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype) would be more expressive in this case.

ntseng (approved):
Looks like it won't break anything. If you end up merging, don't forget to change the commit title.

Issue #419, opened 2025-03-18
Convert bounty board buttons into select


ntseng:
Some bounty interactions were cut because there were more than 5, but we can fit those in by converting some or all bounty board buttons in to a select.

ntseng:
Use patterns in scripts to write sequelize-cli migrations as a new type: "frontend migrations". Also, make new functionality that per guild refreshes all bounty board threads, which can be an admin utility slash command as well.

PR #531, opened 2025-08-02
`/bounty swap` improvements

ntseng:
Summary
-------
Some of the functions being moved or replicated in #419 are old enough to benefit from being refactored. I'm going to split the actual refactors out to separate PRs so that #419's PR can be mostly structural.

Here are the changes for `/bounty swap`:
- add validation for source bounty being completed before slot is selected
- expand destination bounty validation to include destination bounty being completed before slot is selected
- reduce duplicate fetches (allowing the deprecation of `logicLayer.bounties.bulkFindOpenBounties()`)
- switch to `Model.update()` over set and save
- use formating function instead of literal markdown
- avoid recalculating poster level for each bounty when generating slot options

Local Tests Performed
---------------------
- [x] added/expanded validations are needed
- [x] stale bounty detection skips logic if bounty was completed before destination slot was selected
- [x] swaps with open bounties still work
- [x] if bounty in selected slot is completed before slot is selected, bounty in destination slot is not modified

### DIFF
--- a/source/frontend/commands/bounty/swap.js
+++ b/source/frontend/commands/bounty/swap.js
@@ -29,7 +29,8 @@ module.exports = new SubcommandWrapper("swap", "Move one of your bounties to ano
 				collector.on("collect", async (collectedInteraction) => {
 					let previousBounty = openBounties.find(bounty => bounty.id === collectedInteraction.values[0]);
 					if (collectedInteraction.customId.endsWith("bounty")) {
-						const bountySlotCount = Hunter.getBountySlotCount(origin.hunter.getLevel(origin.company.xpCoefficient), origin.company.maxSimBounties);
+						const startingPosterLevel = origin.hunter.getLevel(origin.company.xpCoefficient);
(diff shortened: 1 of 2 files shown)


ntseng:
Convention discussion: This PR is using `let` for some database entities so that the variable can be reassigned with the uptodate entity after `Model.update()`. Alternatively, those entities could remain `const`s, but that would require the use of more `Model.reload()` (which feels likely to create superstitious patterns) or "baton pass"ing the updated variable into a new `const` (which feels likely to cause mixups). While the risk of reassigning the variable opens the possibility of more problems, I think the brevity and minimzation of concepts is more valuable in this case.

ntseng:
Convention discussion: This code includes "double equals" comparisons. I left them alone presuming they were working before, but felt it worth noting that they had an intimidating aura. "I don't remember if I'm suppose to expect heterogeneous types in this comparison, or if this was a 'just in case' usage. Further, even though I've learned it 4-5 times by now, the nuances of how == works feels extra hard to retain."

Issue #537, opened 2025-09-06
Reaction Toasts🥂


ntseng:
Allow users to toast discord messages with auto-generated toast content by reacting with 🥂. Include an option to disable in `/server-config`. Add a randomized tip to explain the feature (consider dynamically removing this tip from the pool in companies that have the option disabled).

ntseng:
- `Message.reply()` to give toast context.
- Additional reactions should be Secondings rather than their own toasts

ntseng:
We should investigate how difficult it'd be to have the reaction toast emoji be customizable per guild.

PR #590, opened 2026-02-10
refactor toast and bounty embed formatters to accept goal data instead of adding later

ntseng:
Summary
-------
- avoid needing "Message Content" intent for Seconding by generating new embed instead of mutating existing embed
- consolidate Server Goal field generation to formatters
- refactor functions that preiously only accepted data from findLatestGoalProgress to also accept data from progressGoal, culling logicLayer hits in several flows
- found bug #589, but split into separate PR except for one line that was hard to untangle

Local Tests Performed
---------------------


ntseng:
Split from work on #537

PR #592, opened 2026-02-16
#537: reaction toasts

### LINKED ISSUE #537: Reaction Toasts🥂

ntseng:
Summary
-------
- add reaction toasts

Local Tests Performed
---------------------
- [x] reacting 🥂 to a message raises a toast to that message's author with the a default message
- [x] reacting 🥂 to a message that already has a reaction toast seconds that reaction toast
- [x] seconding toasts normally still works after refactor
- [x] goal completion embed is shown if reaction toast completes a goal

Issue
-----
Closes #537

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -248,8 +248,8 @@ dAPIClient.on(Events.MessageReactionAdd, async (reaction, user) => {
 
 	// If receiving a Partial, fetch entities
 	let guild = reaction.message.guild;
-	let hostChannel = reaction.message?.channel;
 	let hostMessage = reaction.message;
+	let hostChannel = hostMessage?.channel;
 	if (reaction.partial) {
 		guild = await dAPIClient.guilds.fetch(reaction.message.guildId);
 		hostChannel = await guild.channels.fetch(reaction.message.channelId);


source/index.js
```diff
@@ -234,6 +235,133 @@ dAPIClient.on(Events.InteractionCreate, async interaction => {
...
+	if (reaction.emoji.name !== "🥂") {
+		return;
+	}
+
+	// Reject DM reactions (which lack `.message.guild` and `.message.guildId`), but keep Partials (which lack `.message.guild`)
+	if (!reaction.message.guildId) {
+		return;
+	}
+
+	// If receiving a Partial, fetch entities
```
SillySalamndr: This feels like it belongs to the `if` block a few lines down, and not here quite yet. Is this trying to refer to the defaults used below? It is hard to tell.

source/index.js
```diff
@@ -234,6 +235,133 @@ dAPIClient.on(Events.InteractionCreate, async interaction => {
...
+	if (existingToast && await logicBlob.toasts.wasAlreadySeconded(existingToast.id, user.id)) {
+		return;
+	}
+
+	const previousCompanyLevel = Company.getLevel(company.getXP(await logicBlob.hunters.getCompanyHunterMap(guild.id)));
+	const [season] = await logicBlob.seasons.findOrCreateCurrentSeason(guild.id);
+	const descendingRanks = await logicBlob.ranks.findAllRanks(guild.id);
+	const guildRoles = await guild.roles.fetch();
+	let goalProgress;
+	if (existingToast) {
```
SillySalamndr: I understand why we need a split here, but we could move the code within the `if` and `else` blocks into their own `frontend` calls to better match the logical flow we have for "proper" commands. (Doing so will also help alleviate the cognitive load of this pretty massive function.)

source/index.js
ntseng: In a Partial `reaction.message` is `undefined`, so trying to get the values from the non-Partial path is part of handling Partials.

PR #594, opened 2026-02-18
#462: add GP Festival

### LINKED ISSUE #462: GP Festival
Add commands (maybe under the existing `festival` family) that allows bot managers to toggle on and off a state that multiplies GP contributions.

ntseng:
Summary
-------
- add `/festival start-gp` and `/festival close-gp` for GP Multiplier Festivals
- rename old festivals to `/festival start-xp` and `/festival close-xp`
- add  `nickname` option to `/config-premium` for retaining nicknames after festivals

Local Tests Performed
---------------------
- [x] GP festivals start, end, and multiply gp contributions
- [x] XP festivals still start, end, and multiply xp after refactor
- [x] having a set nickname retains bot nickname after festival

Issue
-----
Closes #462

### DIFF
--- a/source/frontend/shared/dAPIRequests.js
+++ b/source/frontend/shared/dAPIRequests.js
@@ -204,10 +204,10 @@ async function updateBotNicknameForFestival(bountyBotGuildMember, company) {
 	}
 
 	if (tagComponents.length > 0) {
-		const multiplierTag = ` [${tagComponents.map(([type, multiplier]) => `${type} x ${multiplier}`).join(" & ")}]`;
+		const multiplierTag = tagComponents.map(([type, multiplier]) => `${type} x ${multiplier}`).join(" & ");
 		const previousNickname = company.nickname ?? "BountyBot";
 		if (previousNickname.length + multiplierTag.length <= GuildMemberLimits.MaximumDisplayNameLength) {
-			bountyBotGuildMember.setNickname(`${previousNickname}${multiplierTag}`);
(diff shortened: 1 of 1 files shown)


source/frontend/shared/dAPIRequests.js
```diff
@@ -189,6 +190,31 @@ async function unarchiveAndUnlockThread(thread, auditLogReason) {
...
+	}
+	if (company.gpFestivalMultiplier > 1) {
+		tagComponents.push(["GP", company.gpFestivalMultiplier]);
+	}
+
+	if (tagComponents.length > 0) {
+		const multiplierTag = ` [${tagComponents.map(([type, multiplier]) => `${type} x ${multiplier}`).join(" & ")}]`;
+		const previousNickname = company.nickname ?? "BountyBot";
+		if (previousNickname.length + multiplierTag.length <= GuildMemberLimits.MaximumDisplayNameLength) {
+			bountyBotGuildMember.setNickname(`${previousNickname}${multiplierTag}`);
```
SillySalamndr: I'm a bit concerned with the spacing here. Wouldn't this render as something like `BigBountyMachineXP x 1000`? We probably want some idea of spacing between the nickname and the multiplier tag.

ntseng: `multiplierTag` has a whitespace (and square brackets) in it, but yeah that's not what one would expect given the names. Will update to have those in the template string instead.

PR #598, opened 2026-02-20
#568: provide feedback if /moderation xp-penalty is used on banned hunter

### LINKED ISSUE #568: Provide feedback if using `/moderation xp-penalty` on a banned user
The command should notify the moderator that the user is banned (to verify with the moderator's tracking of the banned state), but should not prevent the penalty from being applied. A moderation team may want to ban a user to stop immediate exploits, but need more time to discuss appropriate penalty.

ntseng:
Summary
-------
- provide feedback if /moderation xp-penalty is used on banned hunter
- reduced reliance on potentially partial objects

Local Tests Performed
---------------------
- [x] message is updated if hunter is banned

Issue
-----
Closes #568

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,9 @@
 # BountyBot Change Log
 ## BountyBot Version 2.11.0bfi:
+### Reaction Toasts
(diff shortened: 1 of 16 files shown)


ntseng:
Is the penalized hunter's banned status really relevant to this action? If so, shouldn't most moderation responses include a banned indicator?

PR #654, opened 2026-05-04
#564: add `/bounty ping` command and control panel option

### LINKED ISSUE #564: Command: `/bounty ping`
Mentions (users who've reacted to the bounty's thread + users who are interested in the bounty's event + interal tracking?) with a customizable message.

ntseng:
Summary
-------
- add `/bounty ping` command and control panel option

NOTE: intended for v2.12.0fi (do not merge until after v2.11.1ib ships)

Local Tests Performed
---------------------
- [x] `/bounty ping` mentions users that have marked interested on a bounty's event or reacted to the bounty's thread
- [x] user mentions aren't duplicated if a user has both reacted to the thread and marked interested
- [x] users marked as excluded aren't mentioned

Issue
-----
Closes #564

### DIFF
--- a/source/frontend/shared/flows/bountyPing.js
+++ b/source/frontend/shared/flows/bountyPing.js
@@ -1,30 +1,18 @@
-const { ModalBuilder, LabelBuilder, StringSelectMenuBuilder, UserSelectMenuBuilder, MessageFlags, TextInputBuilder, TextInputStyle, userMention, ModalSubmitInteraction } = require("discord.js");
-const { selectOptionsFromBounties, butIgnoreInteractionCollectorErrors } = require("../../shared");
-const { timeConversion } = require("../../../shared");
-const { SKIP_INTERACTION_HANDLING } = require("../../../constants");
-const { InteractionOrigin } = require("../../classes");
+const { MessageFlags, userMention, ModalSubmitInteraction } = require("discord.js");
 const { Bounty } = require("../../../database/models");
 
 /**
- * @param {typeof import("../../../logic")} logicLayer
  * @param {ModalSubmitInteraction} modalSubmission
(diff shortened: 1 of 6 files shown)


source/frontend/shared/flows/bountyPing.js
```diff
@@ -0,0 +1,51 @@
...
+			}
+		}
+	}
+
+	const excludedHuntersCollection = modalSubmission.fields.getSelectedUsers(labelIds.excludedBountyHunters);
+	if (excludedHuntersCollection) {
+		for (const id of excludedHuntersCollection.keys()) {
+			interestedHunterIds.delete(id);
+		}
+	}
```
SillySalamndr, line 44: Since `interestedHunterIds` is already a proper `Set`, it might be reasonable to use the `difference` function to do some of this work. (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/difference)
```suggestion
	const excludedHuntersCollection = modalSubmission.fields.getSelectedUsers(labelIds.excludedBountyHunters);
	if (excludedHuntersCollection) {
		interestedHunterIds = interestedHunterIds.difference(excludedHuntersCollection.keys());
	}
```

SillySalamndr (commented):
A potential refactor suggestion, since the code appears to be doing a set operation manually. If that refactor doesn't work reasonably, that is understandable.

source/frontend/shared/flows/bountyPing.js
ntseng: Unfortunately, the set comparison methods (`Set.difference()`, `Set.symmetricDifference()`, `Set.intersection()`, etc) don't appear to be in Node yet.

SillySalamndr (approved):

PR #656, opened 2026-05-04
#655: fix `/bounty take-down` not canceling associated Scheduled Event

### LINKED ISSUE #655: `/bounty take-down` doesn't cancel associated Scheduled Event

ntseng:
Summary
-------
- fix `/bounty take-down` not canceling associated Scheduled Event (applied in command, on control panel, and in evergreen)
- consolidate common logic between command and control panel option
- fixed fetches where discord.js was previously returning a collection of 1 event when fetching by id

Local Tests Performed
---------------------
- [x] `/bounty take-down` (command) removes associated Scheduled Event
- [x] bounty control panel take down removes associated Scheduled Event

Issue
-----
Closes #655

### DIFF
--- a/source/frontend/shared/flows/bountyTakeDown.js
+++ b/source/frontend/shared/flows/bountyTakeDown.js
@@ -1,0 +1,33 @@
+const { MessageFlags, Guild } = require("discord.js");
+const { Bounty, Hunter } = require("../../../database/models");
+const { butIgnoreUnknownChannelErrors, butIgnoreMissingPermissionErrors } = require("../dAPIResponses");
+const { syncRankRoles } = require("../dAPIRequests");
+
+/**
+ * @param {typeof import("../../../logic")} logicLayer
+ * @param {Guild} guild
+ * @param {Bounty} bounty
+ * @param {Hunter} posterHunter
+ * @param {import("discord.js").ForumThreadChannel | undefined} bountyThread
+ */
+async function bountyTakeDown(logicLayer, guild, bounty, posterHunter, bountyThread) {
+	await logicLayer.bounties.deleteBountyCompletions(bounty.id);
+	if (bountyThread) {
+		bountyThread.delete("Bounty taken down by poster").catch(butIgnoreMissingPermissionErrors);
+	}
+	if (bounty.scheduledEventId) {
+		guild.scheduledEvents.delete(bounty.scheduledEventId).catch(butIgnoreMissingPermissionErrors);
+	}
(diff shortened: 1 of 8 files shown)


source/frontend/shared/flows/bountyTakeDown.js
SillySalamndr, line 1: I like the directory name for this new pattern.

source/frontend/shared/flows/bountyTakeDown.js
```diff
@@ -0,0 +1,33 @@
...
+		guild.scheduledEvents.delete(bounty.scheduledEventId).catch(butIgnoreErrorIf(isUnknownGuildScheduledEventError, isMissingPermissionError));
+	}
+	bounty.destroy();
+
+	posterHunter.decrement("xp");
+	const [season] = await logicLayer.seasons.findOrCreateCurrentSeason(bounty.companyId);
+	await logicLayer.seasons.changeSeasonXP(bounty.userId, bounty.companyId, season.id, -1);
+	const descendingRanks = await logicLayer.ranks.findAllRanks(bounty.companyId);
+	const seasonalHunterReceipts = await logicLayer.seasons.updatePlacementsAndRanks(await logicLayer.seasons.getParticipationMap(season.id), descendingRanks, await guild.roles.fetch());
+	syncRankRoles(seasonalHunterReceipts, descendingRanks, guild.members);
```
SillySalamndr: In case we ever want/need to wait for `bountyTakeDown` to finish elsewhere, we likely want to `await` the `syncRankRoles` call. If we are in a case where we don't want or need to wait for this `syncRankRoles`, we are likely in a case where we don't need to `await` the rest of this function.

SillySalamndr (commented):
I like this overall. I think there's a good reason to `await` the `syncRankRoles` in the new function, but I can see/understand if you have an argument otherwise.

source/frontend/shared/flows/bountyTakeDown.js
ntseng: Hmm, haven't examined things from a concurrency angle on this yet. I guess I'll return `syncRankRoles`'s `Promise<void>` for now as a way of knowing when that's done in the surrounding context, but we might need to revisit. The way I can think of to get 2 `syncRankRoles` running at the same time would be to have two different shards resolve a flow at the same time, in which case I would definitely expect a race condition of some kind, however we'd probably need more built out locking or something to really address that.

SillySalamndr (approved):

PR #658, opened 2026-05-06
seconding fixes

ntseng:
Summary
-------
- fix seconding a toast removing seconder from recipients on embed
- `toastEmbed` takes array of ids instead of set of ids (most flows submit an array of a single id, set deduping only used in one flow)
- fix reward summary early out not catching no reciept state (companyReciept guaranteed to include 1 or more keys)

Local Tests Performed
---------------------
- [x] seconding a toast no longer removes the seconders' name from the recipient list, but still doesn't grant them xp

### DIFF
--- a/source/frontend/buttons/secondtoast.js
+++ b/source/frontend/buttons/secondtoast.js
@@ -29,14 +29,9 @@ module.exports = new ButtonWrapper(mainId, 3000,
 		const [season] = await logicLayer.seasons.findOrCreateCurrentSeason(interaction.guild.id);
 		const previousCompanyLevel = Company.getLevel(origin.company.getXP(await logicLayer.hunters.getCompanyHunterMap(interaction.guild.id)));
 
-		const recipientIds = [];
-		originalToast.Recipients.forEach(reciept => {
-			if (reciept.recipientId !== interaction.user.id) {
-				recipientIds.push(reciept.recipientId);
-			}
-		});
+		const recipientIds = originalToast.Recipients.map(receipt => receipt.recipientId);
 
-		const hunterReceipts = await logicLayer.toasts.secondToast(origin.hunter, originalToast, origin.company, recipientIds, season.id);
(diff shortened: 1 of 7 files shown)


source/frontend/buttons/secondtoast.js
```diff
@@ -29,14 +29,9 @@ module.exports = new ButtonWrapper(mainId, 3000,
...
-		const recipientIds = [];
-		originalToast.Recipients.forEach(reciept => {
-			if (reciept.recipientId !== interaction.user.id) {
-				recipientIds.push(reciept.recipientId);
-			}
-		});
+		const recipientIds = originalToast.Recipients.map(receipt => receipt.recipientId);
 
-		const hunterReceipts = await logicLayer.toasts.secondToast(origin.hunter, originalToast, origin.company, recipientIds, season.id);
+		const hunterReceipts = await logicLayer.toasts.secondToast(origin.hunter, originalToast, origin.company, recipientIds.filter(id => id !== interaction.user.id), season.id);
```
SillySalamndr, line 34: What is the reasoning for putting the filter on line 34 instead of line 32? Having it happen on line 32 both better matches the previous behavior and better matches the variable name (by already having the set limited to the real recipients).

SillySalamndr (commented):
One possible nitpick? More on the side of flow rather than functionality; everything here looks like it should work.

source/frontend/buttons/secondtoast.js
ntseng: This was actually the bug: the seconder should be removed from the list of recipients for the logic layer (so they don't credit themselves), but not from the list of recipients for the toast embed formatter below.

I chose to make the subset here rather than remake the superset below.

SillySalamndr: Ah, that makes sense. Thank you!

SillySalamndr (approved):

PR #665, opened 2026-05-12
#609 & #561: consolidate thread handling behavior

### LINKED ISSUE #561: Create error filter for "could not post in archived thread"
Although we try to unarchive and unlock threads before posting in them, BountyBot may lack permissions to do so, so it should ignore them if encountered afterwards (error code 50083).

### LINKED ISSUE #609: Additional places to implement `addLogMessageToBountyThread`

ntseng:
Summary
-------
This initially looked like it would be a logic DRYing effort, but context differences made it feel more important to value composibility despite the appearance of duplication.

The main "similar appearing" code has 4 parts, 3 of which vary based on context:
1. Thread mutations (some flows only want to update the embeds, others also want to update the thread name)
2. Attempt to prepare the thread for 3 & 4
3. Acknowledge the pending interaction (reply vs update, ephemeral vs public, etc)
4. Other side-effects (announcements in other channels, etc)

Many of these steps are single lines (after getting required resources [which is also context dependent] and actually getting into the then block that allows them to fire), so creating a larger controller to handle the dynamism feels quite overkill to me at this time.

More important in my opinion is the 4th bullet below: wherein we can make help server admins by keeping their Audit Logs readable.

- cull duplicate fetches of bounty threads
- implement bounty board thread getter to consistently handle thread and forum request misses
- refactored log message logic to reduce duplicate detection logic and avoid awaiting requests known to be missing permissions
- discovered Discord "audit log bundling", and started "one audit log reason per flow" convention

Local Tests Performed
---------------------
- [x] `/bounty record-turn-ins` completes critical path without crashing

Issue
-----
Closes #609, Closes #561

### DIFF
--- a/source/frontend/shared/dAPIRequests.js
+++ b/source/frontend/shared/dAPIRequests.js
@@ -69,11 +69,6 @@ async function unarchiveAndUnlockThread(thread, auditLogReason) {
 	}
 }
 
-/** @param {ThreadChannel} thread */
-function threadCanRecieveMessages(thread) {
-	return !(thread.archived || thread.locked);
-}
-
 /** Updates the embeds in a forum thread's title and starter message
  * @param {Message} starterMessage
  * @param {{ title: string; embed: EmbedBuilder; }} changes
@@ -237,7 +232,6 @@ module.exports = {
 	refreshEvergreenBountiesThread,
 	makeEvergreenBountiesThread,
 	unarchiveAndUnlockThread,
-	threadCanRecieveMessages,
 	refreshBountyBoardThread,
 	getBountyBoardThread,
 	refreshReferenceChannelScoreboardSeasonal,
--- a/source/frontend/shared/dAPIRequests.js
+++ b/source/frontend/shared/dAPIRequests.js
@@ -1,8 +1,9 @@
-const { GuildTextThreadManager, EmbedBuilder, Guild, MessageFlags, Message, GuildMemberManager, ForumChannel, ThreadChannel, GuildMember, GuildScheduledEvent } = require("discord.js");
+const { GuildTextThreadManager, EmbedBuilder, Guild, MessageFlags, Message, GuildMemberManager, ForumChannel, ThreadChannel, GuildMember } = require("discord.js");
 const { Bounty, Company, Rank, Participation } = require("../../database/models");
 const { bountyEmbed, overallScoreboardEmbed, seasonalScoreboardEmbed } = require("./dAPISerializers");
 const { ascendingByProperty } = require("../../shared");
 const { GuildMemberLimits } = require("@sapphire/discord.js-utilities");
+const { butIgnoreUnknownChannelErrors, butIgnoreUnknownMessageErrors } = require("./dAPIResponses");
 
 /**
  * @file Discord API (dAPI) Requests - groups of requests to dAPI formalized into functions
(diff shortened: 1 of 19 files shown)


source/frontend/shared/dAPIRequests.js
```diff
@@ -55,48 +56,56 @@ async function refreshEvergreenBountiesThread(bountyBoardChannel, evergreenBount
...
+ * @param {{ title: string; embed: EmbedBuilder; }} changes
+ * @param {string} auditLogReason the reason to group all changes under in the server's audit log
+ */
+async function refreshBountyBoardThread(starterMessage, { title, embed }, auditLogReason) {
+	if (title !== starterMessage.channel.name) {
+		starterMessage.channel.edit({ name: title, reason: auditLogReason });
+	}
+
+	const starterMessageEditPayload = { embeds: [embed] };
+	if (auditLogReason === "bounty marked completed by poster") {
```
SillySalamndr: This is a pretty long/hefty string to be checking against here, to the point where I'd expect this string to be mistyped regularly. Having this string be a constant somewhere should help reduce the potential for human error.

source/frontend/shared/dAPIRequests.js
```diff
@@ -55,48 +56,56 @@ async function refreshEvergreenBountiesThread(bountyBoardChannel, evergreenBount
...
 	}
-	const bountyBoard = await guild.channels.fetch(company.bountyBoardId);
-	const thread = await bountyBoard.threads.fetch(bounty.postingId);
-	await unarchiveAndUnlockThread(thread, "Unarchived to update posting");
-	return thread.send({ content: auditMessage, flags: MessageFlags.SuppressNotifications });
+	const bountyBoard = await guild.channels.fetch(bountyBoardId).catch(butIgnoreUnknownChannelErrors);
+	if (!bountyBoard) {
+		return null;
+	}
+	return bountyBoard.threads.fetch(postingId).catch(butIgnoreUnknownMessageErrors) ?? null;
```
SillySalamndr: `fetch` returns a Promise, which will always be non-null, even if the `fetch` result will eventually be null. You likely don't need the nullish coalesce here, as `fetch` returns a Promise-wrapped null when it can't find something, anyway.

source/frontend/shared/dAPIRequests.js
SillySalamndr: This string in particular being cross-file makes it stand out more than your other, similar strings that aren't needing to be checked here.

SillySalamndr (changes requested):
I think there's two places where intent could be made a bit clearer (and future mistakes avoided), but they shouldn't be too significant to update each.

SillySalamndr (approved):

PR #695, opened 2026-08-06
item drop fixes

ntseng:
Summary
-------
- fix drops always rolling on most common pool
- fix Unidentified Item and Loot Box consuming Item Find Boost
- remove redundant data from return of `rollItemForHunter`

Local Tests Performed
---------------------


SillySalamndr (approved):
I like the new `createRandomItem` enough that I kind of wish that `rollItemForHunter` were also this clean and/or they shared a code path. That can be a future update, though.

